PR #1, opened 2016-07-07
TMDM-9707 Remain page size setting of foreign key window.

yjli-talend:


qcui:
This change will make ForeignKeyListWindow for each FK sharing the same pageSize setting.

qcui:
Validated.

pbabin:
isPagingAccurate method was introduced when using Qizx. It shouldn't be there any more.

yjli-talend:
Method PagingToolBarEx(int pageSize) still be called in browserecords project,so it can't be made of private.

PR #4, opened 2016-07-08
bugfix/6.3/tmdm 9703 MDM TMDM-9703   tMDMInput with one filter return different records size by different batch size

htyin:


glzhou:
OK to merge

PR #11, opened 2016-07-14
TMDM-9366 Fixed valid don't work in multiple occurence element.

yjli-talend:


glzhou:
missing junit test

yjli-talend:
This junit test is difficult to complete in client side and it is better to use auto scenario test.

zfkang-talend:
Validated.

PR #12, opened 2016-07-15
TMDM-9752 let the FK picker dialog resizable

htyin:


org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/client/widget/ForeignKey/ForeignKeyListWindow.java
```diff
@@ -483,4 +490,9 @@ protected void closeOrHideWindow() {
         return this.typeComboBox;
     }
 
+    private void recalculate(int height, int width) {
+        this.getWidget(0).setHeight(height - 40 + "px"); //$NON-NLS-1$
```
glzhou: It's better not to use magic number, use constant/local variable with a meaningful name instead: i.e: marginSomething

glzhou:
OK to merge

PR #13, opened 2016-07-15
TMDM-9736 Handle primary key info in gwt client side.

yjli-talend:


yjli-talend:
Resolved conflict in local branch.

PR #19, opened 2016-07-19
TMDM-9657 Button In and Out be disabled state when select node is collapsed (in light blue) state.

yjli-talend:


glzhou:
OK to merge

PR #21, opened 2016-07-20
TMDM-9644 Recreate tables of entity dose not delete the FK values dependency to it and the journal records are deleted.

htyin:


glzhou:
OK to merge

PR #23, opened 2016-07-20
TMDM-9095 : Fix the issue, Setting empty value to multi_lingual field…

qcui:
… when fill field value without multi_lingual editor.


org.talend.mdm.webapp.base/src/main/java/org/talend/mdm/webapp/base/client/widget/MultiLanguageField.java
```diff
@@ -202,6 +211,14 @@ protected void doOnFailure(Throwable caught) {
             
         });
     }
+    
+    @Override
+    public boolean validateValue(String value) {
+        if (value != null && !"".equals(value.trim())) { //$NON-NLS-1$
```
htyin: value.trim().isEmpty()  is better.

qcui: "value.trim().isEmpty() is better."
Could you tell me the reason?

htyin:
Validated

PR #32, opened 2016-07-28
TMDM-9804: AUTO_INCREMENT issue in cluster enviroment

zfkang-talend:


org.talend.mdm.core.storage.sql/src/test/java/com/amalto/core/save/DocumentSaveTest.java
```diff
@@ -1748,6 +1749,36 @@ public void testUpdateAutoIncrementRecord() throws Exception {
         assertEquals("Product Family #1", evaluate(committedElement, "/ProductFamily/Name"));
     }
 
+    public void testUpdateAutoIncrementRecordForLongTransactionInCluster() throws Exception {
```
glzhou: Would be good to attach JIRA ID to the junit test case

org.talend.mdm.core/src/com/amalto/core/save/generator/StorageAutoIncrementGenerator.java
```diff
@@ -63,8 +63,12 @@ public StorageAutoIncrementGenerator(Storage storage) {
         this.system = storage;
         // Create auto increment if doesn't exist
         MetadataRepository repository = system.getMetadataRepository();
-        ComplexTypeMetadata autoIncrementType = repository.getComplexType(AUTO_INCREMENT); 
+        ComplexTypeMetadata autoIncrementType = repository.getComplexType(AUTO_INCREMENT);
         UserQueryBuilder qb = from(autoIncrementType).forUpdate();
+
```
glzhou: Can you add some comment about the modification as the code is not self-explained.

PR #35, opened 2016-08-01
TMDM-9824 Add basic mass update using Rest API.

yjli-talend:


pbabin:
Header for new files are missing

PR #36, opened 2016-08-02
TMDM-9774 Forbid Special characters in id element.

zfkang-talend:


qcui:
Save.java line 41
The exception message 'valid character' should be 'invalid character' .

qcui:
Validated

PR #40, opened 2016-08-09
TMDM-9845 Can not create FK record from button "create a new record for this relation" in main tab

htyin:


zfkang-talend:
Validated.

PR #43, opened 2016-08-11
TMDM-9868 While trying to export data from MDM webui to excel file, All data not getting exported.

zfkang-talend:


org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/server/servlet/DownloadData.java
```diff
@@ -153,22 +166,82 @@ protected void fillHeader(HSSFRow row) {
 
     protected void fillSheet(HSSFSheet sheet) throws Exception {
         entity = org.talend.mdm.webapp.browserecords.server.util.CommonUtil.getEntityModel(concept, language);
+
+        Properties mdmConfig = MDMConfiguration.getConfiguration();
+        Object value = mdmConfig.get("max.export.browserecord"); //$NON-NLS-1$
+        if (value != null) {
+            maxCount = Integer.parseInt(value.toString());
+        }
+
```
glzhou: Can we have the parameter initialization in a static block, as it only needs be initialized once.

PR #44, opened 2016-08-11
TMDM-9915 Keep look up field original value when update other fields.

yjli-talend:


org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/client/handler/ItemTreeHandler.java
```diff
@@ -293,4 +301,13 @@ private boolean isEmptyValueEl(Element el) {
         }
         return els;
     }
+    
+    protected String getLookUpFieldValue(ItemBean itemBean,String path, String elValue) {
+    	if (itemBean != null && itemBean.getOriginalLookupFieldMap() != null && itemBean.getOriginalLookupFieldMap().keySet().contains(path)) {
```
htyin: keySet().contains(path)  can be replaced by containsKey(path)

htyin:
Validated

PR #49, opened 2016-08-16
TMDM-9854 : Empty row that next to excel header caused error while importing excel with partial update

qcui:


glzhou:
OK to merge

PR #54, opened 2016-08-23
TMDM-8268 : It's better if we could prompt the line number of the imported record which doesn't satisfy the column definition

qcui:


glzhou:
OK to merge

PR #56, opened 2016-08-26
TMDM-9510 Can't edit Date type element on grid when set display format

zfkang-talend:


org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/client/creator/ItemCreator.java
```diff
@@ -52,7 +52,7 @@ public static ItemBean createDefaultItemBean(String concept, EntityModel entityM
 
             if (typeModel.isSimpleType()) {
 
-                if (typeModel.getType().equals(DataTypeConstants.DATE)) {
+                if (typeModel.getType().getBaseTypeName().equals(DataTypeConstants.DATE.getBaseTypeName())) {
```
glzhou: DataTypeConstants.DATE, is already primitive type, no need to getBaseType again

glzhou:
OK to merge

PR #59, opened 2016-08-30
TMDM-9909 Increase the length of a string element should be low impact

zfkang-talend:


org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/MappingGenerator.java
```diff
@@ -658,22 +659,6 @@ private static void addFieldTypeAttribute(FieldMetadata field,
                     length.setValue("1048576"); //$NON-NLS-1$ 1MB CLOB limit for DB2
                     propertyElement.getAttributes().setNamedItem(length);
                 }
-            } else if (fieldType.getData(MetadataRepository.DATA_MAX_LENGTH) != null) {
```
glzhou: Only need to check max_length from supertype including itself).

glzhou:
OK to merge

PR #67, opened 2016-09-12
TMDM-9939 Bulk update on FK tabs

yjli-talend:


org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/client/widget/treedetail/ForeignKeyTablePanel.java
```diff
@@ -195,6 +214,14 @@ public void initContent(final EntityModel entityModel, ItemNodeModel parent, fin
 
         addListener();
 
+        if (isBulkUpdate) {
+            for (int i = 0; i < fkModels.size(); i++) {
+                fkModels.get(i).setMassUpdate(isBulkUpdate);
+            }
+        }
+        if (fkTypeModel.getForeignKeyFilter() != null) {
```
glzhou: In case of foreign key filter, there should be a warning message such as "Editing is disabled for field(s) that has foreign key filte"

glzhou:
OK to merge

PR #69, opened 2016-09-14
TMDM-9856: [Navigator] Only first FK is retrieved when enclosed in an unbounded complex type

zfkang-talend:


zfkang-talend:
Cherry pick form the TMDM-9954, this is only for automation test. create PR with TMDM-9856 and pushed together.

PR #73, opened 2016-09-21
TMDM-10010: Problem to save an empty complex sub-element

zfkang-talend:


org.talend.mdm.core/src/com/amalto/core/save/DOMDocument.java
```diff
@@ -137,7 +139,26 @@ private void clean(Element element, Cleaner cleaner, boolean removeTalendAttribu
...
-            return element == null || !element.hasChildNodes();
+            if (element == null) {
+                return true;
+            }
+            if (element.getAttributes() != null && element.getAttributes().getLength() > 0) {
+                boolean isAllEmpty = false;
+                for (int i = 0; i < element.getAttributes().getLength(); i++) {
+                    if (StringUtils.isBlank(element.getAttributes().item(i).getNodeValue())) {
+                        isAllEmpty = true;
+                    }
```
pbabin: - element.getAttributes() could called once and assigned to a local variable
- loop could be exited as soon as an attribute value is found

zfkang-talend: modified with PR [#79](https://github.com/Talend/tmdm-server-se/pull/79)

PR #85, opened 2016-10-17
TMDM-9030: Fix unstable Junit tests

zfkang-talend:


org.talend.mdm.core.storage.sql/src/test/java/com/amalto/core/query/StorageQueryTest.java
```diff
@@ -4309,6 +4344,7 @@ public void testOrderByFk() throws Exception {
         try {
             try {
                 for (DataRecord record : records) {
+                    LOG.error("DESC ID="+record.get("Id"));
```
glzhou: Why is it error log?

glzhou:
OK to merge

PR #110, opened 2016-11-17
TMDM-10164 Use 'org.apache.xalan.transformer.TransformerIdentityImpl' class instead of 'net.sf.saxon.IdentityTransformer' to resolve performance issue.

yjli-talend:


pbabin:
Also, there's no information about the gain it brings by changing that. Can you fill in JIRA the comparison of performances before and after?

yjli-talend:
We have 3 places to call ‘TransformerFactory.newInstance().newTransformer()’
1 com.amalto.commons.core.utils.XMLUtils#nodeToString
2 org.talend.mdm.db.migration.DBMigrationTool#nodeToString
3 com.amalto.webapp.core.util.Util#nodeToString
The reasons of hard coding in place one are
1 After testing, we find xalan is faster than saxon when insert record to server. And we are not sure that xalan will be faster in any case, such as in place 2 and place 3,because sax and dom are suitable for different documents.
2 Using hard coding is faster than searching implement class in runtime classpath.
We have added test result in jira attachment file "performanceTestResultWithJob.zip anfRunJobTestResult.xlsx

glzhou:
We have 2 JAXP TransformerFactory implementation in our classpath: Saxon (by default using Sax for parsing) & Xalan (by default using Dom for parsing), right now, JAXP classspath search found Saxon first, therefore resulting some performance degrade for XML => String serialization.

Seems Saxon were added to support XSLT 2.0, For those explicitly use Saxon for xml transforming ( to support XSLT 2), we can keep using it.  For other places (there are more places than what yjli has pointed out) we shall analyze which will provide performance gain. For actions that similar to XML <=> String de/serialization, it seems Xalan (dom) is better. For those that need to access partially the xml document, we should use Saxon (Sax).

Also There are many duplication of function nodeToString() & styleDocument() scattered in different project, we need to reude as much as possible.

glzhou:
Looks good to me.

PR #112, opened 2016-11-18
TMDM-10201 Excel Import: Exported file of related entity is empty when entity is inheritance type

lpw82gf:


glzhou:
OK to merge

PR #118, opened 2016-11-22
TMDM-9703 tMDMInput with one filter return different records size by different batch size

zfkang-talend:


zfkang-talend:
, Model contains one-to-many, page size is 50, if record < 1000, this problem can't appear, If the record count > 1000 , use the contidition: getCount() > IN_CLAUSE_MAX(1000) use StandQueryHandler to query, the query result contains the repeat data, the final result count < 50. 
For the Junit, I modified to create 1500 items record.

org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/InClauseOptimization.java
```diff
@@ -95,7 +95,7 @@ public StorageResults visit(Select select) {
             StorageResults records = storage.fetch(qb.getSelect()); // Expects an active transaction here
             // TMDM-7124. Oracle doesn't like > 1000 number of values in 'IN (...)' clause,
             // and too many values in 'IN (...)' clause hurt database performance
-            if (records.getCount() >= IN_CLAUSE_MAX) {
+            if (records.getSize() >= IN_CLAUSE_MAX) {
                 if (LOGGER.isDebugEnabled()) {
                     LOGGER.debug("Too many ids in 'IN()' clause, abort this optimization. Total ids = " + records.getCount()); //$NON-NLS-1$
```
glzhou: Should use getSize() in the log.

glzhou:
OK to merge

PR #121, opened 2016-11-23
TMDM-8798 FK constraint warning when creating a fk on the fly with special character

htyin:


zfkang-talend:
If the foreign key contains a capital letter with '&', eg: store id = 'A&b', when click save will show the tip message.

zfkang-talend:
Validated.

PR #126, opened 2016-11-25
TMDM-9146 The second FK should show FK infos in journal not show the id value.

htyin:


org.talend.mdm.webapp.journal/src/main/java/org/talend/mdm/webapp/journal/server/ForeignKeyInfoTransformer.java
```diff
@@ -68,23 +71,135 @@ public Document transform(MutableDocument document) {
...
+            path += items.get(i).getPath();
+            if (i == currentBranchIndex) {
+                break;
+            }
+            if (i < items.size() - 1) {
+                path += "/"; //$NON-NLS-1$
+            }
+        }
+        return path;
+    }
```
lpw82gf: These 2 methods could be refactored to share same code.

PR #127, opened 2016-11-28
TMDM-10164 Fixed failed junit test.

yjli-talend:


glzhou:
Should not fix junit, should use the right TransformerFactory(Xalan) for testing. Add Xalan as dependency as MDM SE will need it too.

PR #131, opened 2016-11-30
TMDM-10348: Performance optimization

zfkang-talend:


org.talend.mdm.core/src/com/amalto/core/server/MDMEhCacheUtil.java
```diff
@@ -0,0 +1,74 @@
...
+import net.sf.ehcache.config.PersistenceConfiguration.Strategy;
+import net.sf.ehcache.store.MemoryStoreEvictionPolicy;
+
+public class MDMEhCacheUtil {
+    
+    private static final String MDM_CACHE_MANAGER = "mdmCacheManager";
+
+    private static final int DEFAULT_WAIT_TIME_TO_IDLE_SECONDS = 60;
+    
+    private static final int DEFAULT_WAIT_TIME_TO_LIVE_SECONDS = 120;
```
glzhou: specify the unit 'second' in comment.

org.talend.mdm.core/src/com/amalto/core/server/MDMEhCacheUtil.java
```diff
@@ -0,0 +1,74 @@
...
+        EhCacheCacheManager mdmEhcache = MDMContextAccessor.getApplicationContext().getBean(MDM_CACHE_MANAGER,
+                EhCacheCacheManager.class);
+
+        if (mdmEhcache.getCache(cacheName) == null) {
+            // Create a Cache specifying its configuration.
+            Cache cache = new Cache(new CacheConfiguration(cacheName, MAX_ENTRIES_LOCAL_HEAP)
+                    .memoryStoreEvictionPolicy(MemoryStoreEvictionPolicy.LFU).eternal(false)
+                    .timeToLiveSeconds(DEFAULT_WAIT_TIME_TO_LIVE_SECONDS).timeToIdleSeconds(DEFAULT_WAIT_TIME_TO_IDLE_SECONDS)
+                    .diskExpiryThreadIntervalSeconds(0)
+                    .persistence(new PersistenceConfiguration().strategy(Strategy.LOCALTEMPSWAP)));
```
glzhou: disk persistence is not needed.

org.talend.mdm.core.storage.sql/src/test/java/com/amalto/core/save/AutoIncrementTest.java
```diff
@@ -108,8 +110,13 @@
         LOG.info("Master storage prepared");
         
         BeanDelegatorContainer.createInstance();
+
+        MDMContextAccessor context = new MDMContextAccessor();
+        FileSystemXmlApplicationContext fileContext = new FileSystemXmlApplicationContext(AutoIncrementTest.class.getClassLoader().getResource("../server/mdm-context.xml").getPath());
```
pbabin: I don't understand this change. Better to use ClassPathXmlApplicationContext

zfkang-talend:
We use the EhCache to cache the data cluster, trigger and process, the mdmCacheManager was configured  in the spring applicationContext, have some junit cases were relate with data cluster, so added mdm-context.xml file into the test resource.

org.talend.mdm.core/src/com/amalto/core/util/MDMEhCacheUtil.java
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.cache.ehcache.EhCacheCacheManager;
+
+import com.amalto.core.server.MDMContextAccessor;
+
+import net.sf.ehcache.Cache;
+import net.sf.ehcache.Element;
+import net.sf.ehcache.config.CacheConfiguration;
+import net.sf.ehcache.config.PersistenceConfiguration;
+import net.sf.ehcache.config.PersistenceConfiguration.Strategy;
+import net.sf.ehcache.store.MemoryStoreEvictionPolicy;
```
qcui: need to remove unused dependency class

org.talend.mdm.core/src/com/amalto/core/util/MDMEhCacheUtil.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.amalto.core.server.MDMContextAccessor;
+
+import net.sf.ehcache.Element;
+
+public class MDMEhCacheUtil {
+
+    public static final String MDM_CACHE_MANAGER = "mdmCacheManager"; //$NON-NLS-1$
+
+    public static Object getCache(String cacheName, Object key) {
```
pbabin: - Would be better to get rid of this Java6 signature to avoid clients to cast the returned type and use Java7 such:
public static <T> T getCache(String cacheName, Object key, class<T> cls)
- If it's an utility class, then there shouldn't be any constructor

PR #136, opened 2016-12-06
TMDM-8140 Fixed can not edit foreign key record that foreign key info contains another foreign key.

yjli-talend:


qcui:
"return null;" would be lost foreign key display info.

qcui:
Validated.

PR #139, opened 2016-12-07
TMDM-10244 MDM stored procedure results of null values are stripped off

zfkang-talend:


yjli-talend:
Is it possible to use 'DataRecordIncludeNullValueXmlWriter' class to extend 'DataRecordXmlWriter' class in order to avoid some duplicated code.

PR #145, opened 2016-12-13
TMDM-8831 Update 'ItemBeam' object in first tab of detail panel when click 'Save' button in foreign key tab.

yjli-talend:


qcui:
validated

PR #156, opened 2016-12-23
TMDM-7410: Journal previous and next change buttons do not work when change is linked to a reusable type.

zfkang-talend:


qcui:
replaceAll("\\[\\d+\\]", "") already contains "[1]" case

qcui:
validated

PR #157, opened 2016-12-23
TMDM-10429 Using xpath instead of id in column of list.

yjli-talend:


qcui:
This change makes regression issue, couldn't display FK infor field content.

qcui:
validated

PR #173, opened 2017-01-16
MDM REST API does not handle correctly some characters when output is JSON

zfkang-talend:


org.talend.mdm.core/test/com/amalto/core/storage/StorageMetadataUtilsTestCase.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.talend.mdm.commmon.metadata.ComplexTypeMetadata;
+import org.talend.mdm.commmon.metadata.FieldMetadata;
+import org.talend.mdm.commmon.metadata.MetadataRepository;
+
+import com.amalto.core.storage.record.DataRecord;
+import com.amalto.core.storage.record.metadata.DataRecordMetadataImpl;
+
+public class StorageMetadataUtilsTestCase {
```
glzhou: Use **Test for class name as it's the convention.

glzhou:
OK to merge

PR #189, opened 2017-02-04
TMDM-9941 Fixed display change message error in studio when modify entity.

yjli-talend:


org.talend.mdm.core/src/com/amalto/core/storage/services/SystemModels.java
```diff
@@ -251,7 +252,13 @@ public String analyzeModelChange(@ApiParam("Model name") @PathParam("model") Str
             Compare.DiffResults diffResults = Compare.compare(previousRepository, newRepository);
             // Analyzes impacts on the select storage
             impacts = storage.getImpactAnalyzer().analyzeImpacts(diffResults);
-            typesToDrop = storage.findSortedTypesToDrop(diffResults, true);
+            List<ComplexTypeMetadata> typesToDrop = storage.findSortedTypesToDrop(diffResults, true);
+            Set<String> findTablesToDrop = storage.findTablesToDrop(typesToDrop);
```
lpw82gf: **Set<String> findTablesToDrop** better rename to **tableNamesToDrop**, same with **typeNamesToDrop**

PR #191, opened 2017-02-07
TMDM-3612 You can define a data model with cyclic stucture, save, deploy, and this causes the server to enter an infinite loop, eventually ending in StackOverflowError

htyin:


zfkang-talend:
Validated.

PR #197, opened 2017-02-21
TMDM-10622 : Bulk load will execute extra UPDATE SQL for x_talend_tas…

lpw82gf:
…k_id & x_talend_timestamp


glzhou:
OK to merge

PR #203, opened 2017-02-24
TMDM-9173 : Can't use a FK as a Match Rule element, when running integrated matching

qcui:


org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/FlatTypeMapping.java
```diff
@@ -155,7 +155,7 @@ public void setValues(Session session, DataRecord from, Wrapper to) {
                 } else if (field instanceof ReferenceFieldMetadata) {
                     StorageClassLoader storageClassLoader = (StorageClassLoader) Thread.currentThread().getContextClassLoader();
                     if (!field.isMany()) {
-                        if (value != null) {
+                        if (value != null && "class com.amalto.core.storage.record.DataRecord".equals(value.getClass().toString())) { //$NON-NLS-1$
```
glzhou: Do not use text comparation, use instanceof

glzhou:
OK to merge

PR #204, opened 2017-03-01
TMDM-10499 : Any manual change performed on a Master record should be…

lpw82gf:
… propagated to Staging


org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/server/actions/BrowseRecordsAction.java
```diff
@@ -2653,4 +2669,20 @@ private static String readRestErroMessage(HttpResponse response) {
     		return sb.toString();
     	}
 	}
+
+    private String getStagingStatus(WSDataClusterPK wsDataClusterPK, String conceptName, String taskId) {
+        String status = StringUtils.EMPTY;
+        String query = "select max(" + Storage.METADATA_STAGING_STATUS + ") from " + conceptName + " where " //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
```
glzhou: Should use the real logic as whether groupId = XXX and status = 205 record exists in Staging

glzhou:
OK to merge

PR #209, opened 2017-03-06
TMDM-10563 page info is incorrect use fulltext search when records count greater than page size

zfkang-talend:


glzhou:
OK to merge

PR #210, opened 2017-03-07
TMDM-10534: [Impact Analysis] Move a complex mandatory field (containing a simple optional field) to optional

zfkang-talend:


glzhou:
OK to merge

PR #217, opened 2017-03-15
TMDM-9366: Save issue for multiple occurence element

zfkang-talend:


yjli-talend:
This commit can't fix issue that jira describe.

PR #218, opened 2017-03-20
TMDM-9366: Save issue for multiple occurence element

zfkang-talend:


qcui:
Need modify testValidateValue() customer format Test case.

qcui:
validated

PR #232, opened 2017-04-14
TMDM-10491 Propagate changes in MDM Data model onto TDS schema & camp…

htyin:
Fix the regression of TMDM-10534


glzhou:
OK to merge

PR #260, opened 2017-05-22
TMDM-10859 Update request - critical CVEs on dependencies.

yjli-talend:


org.talend.mdm.core/pom.xml
```diff
@@ -13,7 +13,7 @@
     <name>org.talend.mdm.core</name>
 
     <properties>
-        <jackson.version>1.9.2</jackson.version>
+        <jackson.version>1.9.13</jackson.version>
```
glzhou: according to https://stackoverflow.com/questions/38017676/small-fix-for-cve-2016-3720-with-older-versions-of-jackson-all-1-9-11-and-in-ja

CVE-2016-3720 is not backported to 1.9, we should upgrade to 2.7.4 or later.

org.talend.mdm.core/pom.xml
```diff
@@ -150,13 +150,8 @@
...
             <groupId>org.apache.activemq</groupId>
-            <artifactId>activemq-pool</artifactId>
-            <version>5.7.0</version>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.activemq</groupId>
-            <artifactId>activemq-core</artifactId>
-            <version>5.7.0</version>
+            <artifactId>activemq-all</artifactId>
+            <version>5.14.5</version>
```
glzhou: we only take what we need. No need to include all the activemq jars

org.talend.mdm.core/pom.xml
```diff
@@ -176,23 +171,18 @@
...
-        <dependency>
-            <groupId>org.codehaus.jackson</groupId>
-            <artifactId>jackson-mapper-asl</artifactId>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
             <version>${jackson.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.codehaus.jackson</groupId>
-            <artifactId>jackson-jaxrs</artifactId>
```
glzhou: jaxrs jars are splitted into 
http://mvnrepository.com/artifact/com.fasterxml.jackson.jaxrs

PR #268, opened 2017-06-01
TMDM-10985: Unable to alter schema on mysql on LINUX

zfkang-talend:


org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/LiquibaseSchemaAdapter.java
```diff
@@ -144,6 +145,11 @@ public void adapt(Connection connection, Compare.DiffResults diffResults) throws
                     columnName = columnName.toUpperCase();
                 }
 
+
+                if (dataSource.getDialectName() == DataSourceDialect.POSTGRES) {
```
glzhou: Postgres is case insensitive unless you use double quote on table/column names, https://www.postgresql.org/docs/current/static/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
Is it necessary for the toLowerCase on Postgresql DB?

zfkang-talend: If connected liunx postgres server, liquibase gernate sql is not wrap with double quote, how if connected windows prostgres server, liquibase gernate sql is wrap with double quote. 

liquibase log for liunx postgres server

> ALTER TABLE public.product ALTER COLUMN  x_description DROP NOT NULL;

liquibase log for windows postgres server

> ALTER TABLE public."Product" ALTER COLUMN  x_description DROP NOT NULL

glzhou:
LGTM

PR #274, opened 2017-06-05
TMDM-10887 : Having a TDS issue, running validation does not issue an…

lpw82gf:
…y error (fix status code issue)


glzhou:
OK to merge

PR #282, opened 2017-06-13
TMDM-9848 Don't create delete journal event in staging container.

yjli-talend:


org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/server/actions/BrowseRecordsAction.java
```diff
@@ -237,8 +237,7 @@
                     WSDeleteItemWithReport wsDeleteItem = new WSDeleteItemWithReport(new WSItemPK(new WSDataClusterPK(
                             dataClusterPK), concept, ids), UpdateReportPOJO.GENERIC_UI_SOURCE,
                             UpdateReportPOJO.OPERATION_TYPE_PHYSICAL_DELETE, "/", //$NON-NLS-1$
-                            LocalUser.getLocalUser().getUsername(), true, true, override);
-
+                            LocalUser.getLocalUser().getUsername(), true, org.talend.mdm.webapp.browserecords.server.util.CommonUtil.isPushToUpdateReport(dataClusterPK), override);
```
glzhou: What about other clients? Components, Studio etc.  constraint should be set in the backend.

PR #285, opened 2017-06-14
TMDM-11011: Migration:migrate from 542 to 640 master error issues.

zfkang-talend:


glzhou:
No patch has been requested, PR is not needed at this moment.

PR #287, opened 2017-06-16
TMDM-9999 Fixed issue that the 'Look up' field doesn't work when open item in new tab in 'browserecords' and 'hierarchy'.

yjli-talend:


zfkang-talend:
Validated.

PR #291, opened 2017-06-19
TMDM-10993: Unable to redeploy a DataModel after adding a reusable complex type Element to an Entity

zfkang-talend:


org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/StorageTableResolver.java
```diff
@@ -135,7 +140,12 @@ public String getFkConstraintName(ReferenceFieldMetadata referenceField) {
         // with same
         // length but different name.
         if (!referenceFieldNames.add(referenceField.getContainingType().getName().length() + '_' + referenceField.getName())) {
-            return formatSQLName("FK_" + Math.abs(referenceField.getName().hashCode()) + fkIncrement.incrementAndGet()); //$NON-NLS-1$
+            String fkName = "FK_" + Math.abs(referenceField.getName().hashCode()) + fkIncrement.incrementAndGet();
+            if (!fkNameMap.containsKey(referenceField.getEntityTypeName())) {
+                fkName = "FK_" + Constraint.hashedName("table`" + referenceField.getEntityTypeName()+"`column`"+referenceField.getName()+"`");
```
lpw82gf: Need to format the code.

glzhou: - Why do we need introduce dependency of org.hibernate.mapping.Constraint in this method? Can't it be simple hash function?
- A little descriptive comment will help to understand the change here

PR #296, opened 2017-06-28
TMDM-10793 : Predicate definition of where conditions missing after deploy view to mdm server

qcui:


pbabin:
What if I have a unique condition. Seems it will be replaced by a AND right? if so, sounds a bit weird to me

qcui:
If server side keep predicate 'NONE' status, then can keep both 'NONE' and 'AND' predicate status.

PR #298, opened 2017-06-28
TMDM-11072 : Support of AWS ELB for MDM Cluster

lpw82gf:
Using all the cookies got from previous response to build a new request.


org.talend.mdm.bulkload.client/src/org/talend/mdm/bulkload/client/BulkloadClient.java
```diff
@@ -154,12 +155,12 @@ public void setToken(String tokenKey, String tokenValue) {
         this.tokenValue = tokenValue;
     }
 
-    public String getSessionId() {
-        return sessionId;
+    public List<String> getCookie() {
```
pbabin: Can it be renamed to getCookies() ?

PR #303, opened 2017-06-30
TMDM-11023 element of date type with display format should display with same value in grid list and detail panel

zfkang-talend:


qcui:
convertStringToDate(String value) function use catch exception logic to implement dateTimeFormat transformation is not suitable. use more clear logic is better.

org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/server/actions/BrowseRecordsAction.java
```diff
@@ -2464,9 +2465,37 @@ public boolean isExistId(String concept, String[] ids, String language) throws S
...
+            sdf = new SimpleDateFormat(dateFormat, java.util.Locale.ENGLISH);
+        } else {
+            sdf = new SimpleDateFormat(dateTimeFormat, java.util.Locale.ENGLISH);
+        }
+        
+        Date date = null;
+        try {
+            date = sdf.parse(value);
+        } catch (ParseException e) {
+            LOG.debug("model object is not a data type", e);
```
qcui: Throw the exception can indicate real fail reason.

qcui: Throw the exception can indicate real fail reason.

qcui:
Add test case about convertStringToDate fail case.

org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/server/actions/BrowseRecordsAction.java
```diff
@@ -2464,9 +2465,38 @@ public boolean isExistId(String concept, String[] ids, String language) throws S
         }
     }
 
+    private Date convertStringToDate(String value) throws ServiceException {
```
pbabin: Duplication of org.talend.mdm.webapp.base.server.util.DateUtil

zfkang-talend:
Before modify, we used te Date type to transfer the value from client to server, in fact, Date is used a long value to store the milliseconds since January 1, 1970, 00:00:00 GMT. beacuse the time zone of client is different with the server, the server get the value is not same with the client.  we used the string to transfer the value.

org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/server/actions/BrowseRecordsAction.java
```diff
@@ -2467,12 +2468,28 @@ public boolean isExistId(String concept, String[] ids, String language) throws S
...
+            if (dateValue.length() == dateFormat.length()) {
+                d = DateUtil.convertStringToDate(dateFormat, dateValue);
+            } else {
+                d = DateUtil.convertStringToDate(dateTimeFormat, dateValue);
+            }
+            if (d != null) {
+                model.setObject(d);
+            }
+        }
+
```
pbabin: Logic here is really weird

org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/server/actions/BrowseRecordsAction.java
```diff
@@ -213,6 +214,10 @@
 
     public static final String FAIL_KEYWORD = "FAIL";//$NON-NLS-1$
 
+    public static final String DATE_FORMAT = "yyyy-MM-dd"; //$NON-NLS-1$
+
+    public static final String DATE_TIME_FORMAT = "yyyy-MM-dd'T'HH:mm:ss"; //$NON-NLS-1$
```
pbabin: This should be defined into DateUtil

org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/client/widget/inputfield/FormatDateField.java
```diff
@@ -163,10 +163,12 @@ public void formatValue(String value) {
...
     private void setFormatedValue(final FormatValueCallback callback) {
         if (formatPattern != null && formatPattern.trim().length() > 0) {
+            String rawValue = this.getRawValue();
             try {
-                Date d = propertyEditor.convertStringValue(this.getRawValue());
+                Date d = propertyEditor.convertStringValue(rawValue);
                 if (d != null) {
-                    final FormatModel model = new FormatModel(formatPattern, d, Locale.getLanguage());
+                    final FormatModel model = new FormatModel(formatPattern, rawValue, Locale.getLanguage());
+                    model.setDate(true);
```
pbabin: It should not be the client to tell server what type is the field value it is manipulating . Server side should know that. Not sure what you can do in this area without too much changes tho but for sure it would avoid the very odd method getFormat() below to occur on server side to identify from what the client sends, it is a Date or a Datetime

zfkang-talend:
@pbabin 
The fix did the following: 
Problem: Date/Datetime as Object received by server side is already using client side timezone to set epoch value (it should be happening at the client side)
Solution: Instead of sending the Date Object, we send plantext Date/Datetime string with fixed format. So when server parse the date and reformat the date according to user's custom format, it will use server side timezone. Then the formatted string will sent back to client.

PR #308, opened 2017-07-07
TMDM-11110: Match & merge execution failed using Product datamodel(oracle db)

zfkang-talend:


org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/MappingGenerator.java
```diff
@@ -582,7 +582,7 @@ private Element handleSimpleField(FieldMetadata field) {
                 Attr tableName = document.createAttribute("table"); //$NON-NLS-1$
                 tableName.setValue(resolver.getCollectionTable(field));
                 listElement.getAttributes().setNamedItem(tableName);
-                if (field.getContainingType().getKeyFields().size() == 1) {
+                if (field.getContainingType().getKeyFields().size() == 1 && dataSource.getDialectName() != RDBMSDataSource.DataSourceDialect.ORACLE_10G) {
```
glzhou: Don't think changing fetch strategy for Oracle is the right solution

PR #312, opened 2017-07-11
TMDM-10659	 Continued select count(*) queries are slow on large volumes

lpw82gf:
1, Add count data to cache to avoid frequently query, data saved in map, with Key={storageName}#{storageType}.{entityName}/{condition hash}, like `Product#MASTER.Product/3214568`
2, When data deleted/updated/created, after transaction committed, reset cache
3, When data cluster dropped, reset all the count cache belongs to this cluster
4, Standalone server using In Memory approach, while cluster environment using Hazelcast approach, like AutoIncrement.


org.talend.mdm.core.storage.sql/src/test/java/com/amalto/core/query/LongTransactionAbstractTestCase.java
```diff
@@ -38,7 +39,7 @@
     
     @Override
     public void setUp() throws Exception {
-        // override to avoid storage.begin in superclass setup
+        EntityCountUtil.clearAll();
```
pbabin: Can the previous comment be kept. It is important one to understand why we don't call super.setUp

org.talend.mdm.core/src/com/amalto/core/storage/InMemoryCounter.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package com.amalto.core.storage;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.amalto.core.storage.Counter.AbstractCounter;
+
+public class InMemoryCounter extends AbstractCounter implements Counter {
```
pbabin: Both implementations of AbstractCounter are using an internal Map<String, Integer> . Cant' it be then part of the base class, thus simplifying a lot their implementations?

org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/HibernateStorage.java
```diff
@@ -141,6 +141,9 @@
 import com.amalto.core.server.MetadataRepositoryAdmin;
 import com.amalto.core.server.ServerContext;
 import com.amalto.core.server.StorageAdmin;
+import com.amalto.core.storage.Counter.CountKey;
+import com.amalto.core.storage.Counter.CountKeyList;
```
htyin: CountKeyList is never used.

lpw82gf: It is used in `EntityCountUtil#addToClearList`

PR #314, opened 2017-07-13
TMDM-8706 Cancel timer refresh before starting validation in order to avoid that validation and timer will refresh ui in same time.

yjli-talend:


lpw82gf:
LGTM

PR #315, opened 2017-07-19
TMDM-10719: Display format on the grid list is ignored for a date field part of a complex type

zfkang-talend:


qcui:
LGTM

org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/server/util/CommonUtil.java
```diff
@@ -58,6 +66,20 @@
 
     private static final Logger LOG = Logger.getLogger(CommonUtil.class);
 
+    public static final String ORIGINAL_VALUE = "originalValue";
+
+    public static final String FORMATE_VALUE = "formateValue";
```
pbabin: typo, should be FORMATTED_VALUE

PR #339, opened 2017-08-16
TMDM-11181 Using token instead of username and password to login tds server.

yjli-talend:
Using username and password to login tds server.

**What is the new behavior?**
Using token to login tds server.

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:


org.talend.mdm.core/src/com/amalto/core/server/security/SecurityUtils.java
```diff
@@ -94,7 +95,13 @@ public static String getIdToken() {
         Map<String, Object> map = (Map<String, Object>) authentication.getUserAuthentication().getDetails();
         return ID_TOKEN + " " + map.get(ID_TOKEN);
     }
-
+    
+    public static String getToken() {
```
pbabin: This method duplicates code from getCredentials method?

lpw82gf: They are different tokens, here we need **Bearer Token**, and for getCredentials, that is id_token (serialized string by Spring)

PR #345, opened 2017-08-22
TMDM-11334 Multiple issues on WebUI after using SSO

yjli-talend:

**What is the new behavior?**
Issue 1 Relation button doesn't work
After we use WebSecurityConfigurerAdapter in sso,spring add 2 headers in response,they are 
X-Content-Type-Options: nosniff and X-XSS-Protection: 1; mode=block.IE 9(include above 9 version) need to specify Content-Type header in response.Content will be blocked by ie if miss Content-Type or Content-Type dismatched.
Please refer to
https://msdn.microsoft.com/query/dev12.query?appId=Dev12IDEF1&l=EN-US&k=k(VS.WebClient.Help.SEC7112)
https://blogs.msdn.microsoft.com/ie/2010/10/26/mime-handling-changes-in-internet-explorer/
Relation button will call dwr,we need load ItemsbrowserInterface.js from dwr,but dwr return text/plain content-type,it mismatch with javascript type.
Refer to dwr source code in org.directwebremoting.servlet.InterfaceHandler
public void handle(HttpServletRequest request, HttpServletResponse response) throws IOException
    {
        String scriptName = request.getPathInfo();
        scriptName = LocalUtil.replace(scriptName, interfaceHandlerUrl, "");
        scriptName = LocalUtil.replace(scriptName, PathConstants.EXTENSION_JS, "");
        if (!LocalUtil.isJavaIdentifier(scriptName))
        {
            throw new SecurityException("Script names may only contain Java Identifiers");
        }

        String path = request.getContextPath() + request.getServletPath();

        String script = remoter.generateInterfaceScript(scriptName, path);

        // Officially we should use MimeConstants.MIME_JS, but if we cheat and
        // use MimeConstants.MIME_PLAIN then it will be easier to read in a
        // browser window, and will still work just fine.
        response.setContentType(MimeConstants.MIME_PLAIN);
        PrintWriter out = response.getWriter();
        out.print(script);
    }
So update dwr version to 2.0.11-RELEASE

Issue 2 Picture dose not show normal on IE
It has the same reason with issue 1,we add Content-Type header in response.

Issue 3 H2 console dose not work
H2 console accessibility: this is not a regression introduced by IAM. This is something we wanted to prevent for security concerns. If someone really wants to get access to the webapp from a different location, then it has to be customized manually.

Issue 4 Can't redirect to login page
We need to identify different login page after we use IAM

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:


org.talend.mdm.webapp.base/src/main/java/org/talend/mdm/webapp/base/client/SessionAwareAsyncCallback.java
```diff
@@ -72,7 +72,7 @@ private boolean sessionExpired(Throwable caught) {
             // When session expires container will use a configured expired Url to return the content of the login page
             // However, since an GWT RPC response cannot be of HTML text, the InvocationException is thrown with the
             // HTML content as the message.
-            return msg == null ? false : msg.contains("<meta name=\"description\" content=\"Talend MDM login page\"/>"); //$NON-NLS-1$
+            return msg == null ? false : (msg.contains("<meta name=\"description\" content=\"Talend MDM login page\"/>") || msg.contains("<title>Talend - Login</title>")); //$NON-NLS-1$
```
pbabin: Are we enough confident this wouldn't be changed?

org.talend.mdm.ext.images.server/src/main/java/talend/ext/images/server/ImageLocateServlet.java
```diff
@@ -111,4 +119,29 @@ private String getResourceFilePath(HttpServletRequest req) {
         path = path.replaceAll("\\\\", "/"); //$NON-NLS-1$//$NON-NLS-2$
         return path;
     }
+    
+    private String getResourceFileMIMEType(String fileType) {
```
pbabin: This is inconsistent with https://github.com/Talend/tmdm-server-se/blob/e83f44fa571ade0ce95bbbafca2f1940869e6249/org.talend.mdm.webapp.openmdm/src/main/webapp/WEB-INF/web.xml#L331

org.talend.mdm.ext.images.server/src/test/java/talend/ext/images/server/util/ImagePathUtilTest.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (C) 2006-2016 Talend Inc. - www.talend.com
```
pbabin: 2016? Are your eclipse settings out-of date?

org.talend.mdm.ext.images.server/pom.xml
```diff
@@ -35,6 +35,12 @@
             <artifactId>thumbnailator</artifactId>
             <version>0.4.8</version>
         </dependency>
+        <!-- Test dependencies -->
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
```
pbabin: scope test is already defined in base pom, right?

org.talend.mdm.webapp.base/src/main/java/org/talend/mdm/webapp/base/client/SessionAwareAsyncCallback.java
lpw82gf: It is not controlled by us, but IAM should not change that easily. Maybe we have to do that for now.

org.talend.mdm.ext.images.server/src/main/java/talend/ext/images/server/ImageLocateServlet.java
```diff
@@ -35,6 +38,8 @@
     private static final long serialVersionUID = -3012919798771313147L;
 
     private static final Logger LOGGER = Logger.getLogger(ImageLocateServlet.class);
+    
+    private final String CONTENT_TYPE_HEADER = "Content-Type";
```
lpw82gf: Could be `private static final`  ?

org.talend.mdm.ext.images.server/src/main/java/talend/ext/images/server/ImageLocateServlet.java
```diff
@@ -63,6 +68,11 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
                 String strHeight = request.getParameter("height"); //$NON-NLS-1$
                 String strPreserveAspectRatio = request.getParameter("preserveAspectRatio"); //$NON-NLS-1$
 
+                Path source = Paths.get(resourceFilePath);
+                String contentType = Files.probeContentType(source);
```
pbabin: Files#probeContentType default implementation is OS-specific and not very complete. Has it been verified it works as expected on all server OSes we support

yjli-talend: http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8008345,it doesn't work in mac.It is easier to write a method to do that.It is unnecessary to add new dependency as your solution in the list.

pbabin: AFAIK MacOS is not part of the list of supported servers

PR #366, opened 2017-09-13
TMDM-11363: JMS connections issues while reindexing

pbabin:
The ConnectionFactory used with a JMSTemplate object should always return pooled Connections (or a single shared Connection) as well as pooled Sessions and MessageProducers.
- Fix to use of CachingConnectionFactory, this factory switches the “reconnectOnException” property to “true” by default, allowing for automatic recovery of the underlying Connection.
- Consider also the ability to raise the “sessionCacheSize” value in case of a high-concurrency environment using a mdm.conf 'mdm.routing.engine.broker.sessionCacheSize' property.
- Fix incorrect console log message

ConnectionFactory isn't pooled for reindexation replication in Cluster as well as for Staging tasks cancellation.

**What is the new behavior?**
Change bean name (instead of changing code which is getting non-pooled CF) of the pooled CF defined for routing events, so that everywhere the right CF is used.
Give the ability to tune CF when huge JMS traffic.

- [x ] The commit message follows Talend standard

- [ x] Bugfix
- [ ] Feature

- [ ] Yes
- [x ] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/resources/META-INF/mdm-context.xml
+++ b/org.talend.mdm.core/resources/META-INF/mdm-context.xml
@@ -44,23 +44,24 @@ http://activemq.apache.org/schema/core http://activemq.apache.org/schema/core/ac
     <import resource="${mdm.jms.provider.xml:activemq-provider.xml}"/>
     
     <!-- Caching connection factory for JMS template performances -->
-    <bean id="jmsPooledConnectionFactory" class="org.springframework.jms.connection.CachingConnectionFactory">
-        <constructor-arg ref="jmsConnectionFactory"/>
+    <bean id="jmsConnectionFactory" class="org.springframework.jms.connection.CachingConnectionFactory">
+        <property name="targetConnectionFactory" ref="spiJmsConnectionFactory"/>
+        <property name="sessionCacheSize" value="${mdm.routing.engine.broker.sessionCacheSize:1}"/>
     </bean>
     
     <!-- Single connection and several sessions -->
     <bean id="jmsSingleConnectionFactory" class="org.springframework.jms.connection.SingleConnectionFactory">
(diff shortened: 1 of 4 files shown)


org.talend.mdm.core/resources/META-INF/mdm-context.xml
```diff
@@ -44,23 +44,24 @@ http://activemq.apache.org/schema/core http://activemq.apache.org/schema/core/ac
     <import resource="${mdm.jms.provider.xml:activemq-provider.xml}"/>
     
     <!-- Caching connection factory for JMS template performances -->
-    <bean id="jmsPooledConnectionFactory" class="org.springframework.jms.connection.CachingConnectionFactory">
-        <constructor-arg ref="jmsConnectionFactory"/>
+    <bean id="jmsConnectionFactory" class="org.springframework.jms.connection.CachingConnectionFactory">
+        <property name="targetConnectionFactory" ref="spiJmsConnectionFactory"/>
+        <property name="sessionCacheSize" value="${mdm.routing.engine.broker.sessionCacheSize:1}"/>
```
lpw82gf: We don't need to update mdm.conf to set default value for `mdm.routing.engine.broker.sessionCacheSize`?

pbabin: No, as you can see above, default value is 1, so having property in mdm.conf is optional.
It's only if you want to increase value that you can add the property. In that case it will override the default value.

lpw82gf: Then we need DOC for that optional config as it isn't in mdm.conf.

pbabin: Right, that's the purpose of the associated Jira TMDM-11414

PR #373, opened 2017-09-20
TMDM-11228 Provide new operator "IN" for query language

zfkang-talend:
have no the function of the *in* when query data.

**What is the new behavior?**
- add the new function of *in/notIn* when query data.
- if the query condition is multiple condition, include 'in' operation and 'contains' operation, use the database query, can not use full text search.
 

- [ ] Bugfix
- [x] Feature

- [ ] Yes
- [x] No

**Other information**:


org.talend.mdm.core/src/com/amalto/core/query/user/ConstantExpression.java
```diff
@@ -11,11 +11,22 @@
...
 
+    public List<T> getValueList();
+
+    default Object getConstant(){
+        if (getValue() != null && getValueList().isEmpty()) {
+            return getValue();
+        } else {
+            return getValueList();
+        }
+    }
```
pbabin: Could it be possible to avoid having weird condition on the nullity of the value and the emptiness of the values list. Also method names could be consistent such `getValue`, `getValueList`, `getValueObject`, `getId`, `getIdList`, `getIdObject`, ...
Such you would get something like:
```
public interface ConstantExpression<T extends Comparable> extends TypedExpression {
    public T getValue();
    public List<T> getValueList();
    public boolean isExpressionList()
    default Object getValueObject() {
        if (isExpressionList()) {
            return getValue();
        } else {
            return getValueList();
        }
    }
}
```
An implementation could also avoid instantiating an `ArrayList` object for nothing and could easily check illegal calls.
```
public class FloatConstant implements ConstantExpression<Float> {
    private final Float value;
    private final List<Float> valueList;

    public FloatConstant(String value) {
        assert value != null;
        this.value = Float.parseFloat(value);
        this.valueList = null;
    }
    public FloatConstant(List<Float> valueList) {
        assert valueList != null;
        this.value = null;
        this.valueList = valueList;
    }
    public boolean isExpressionList() {
        return this.valueList != null;
    }
    public Float getValue() {
        if (isExpressionList()) {
            throw new IllegalStateException();
        }
        return value;
    }
    public List<Float> getValueList() {
        if (!isExpressionList()) {
            throw new IllegalStateException();
        }
        return valueList;
    }
}
```

org.talend.mdm.core/src/com/amalto/core/query/user/Id.java
```diff
@@ -52,36 +54,47 @@ public String getId() {
         return id;
     }
 
-    public List<String> getConstantCollection() {
-        return constantCollection;
+    public List<String> getIdList() {
```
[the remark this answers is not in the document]
zfkang-talend: If the condition field is a key field or reference field, will used the Id.java to store the id(eq) or idList(in/notIn).  Class Id.java is implements the interface Expression.java, Expression.java can't define abstract method isExpressionList(), isExpressionList() method is defined in interface ConstantExpression.java. In Id.java, used isExpressionList() method in hashcode() and equals() method.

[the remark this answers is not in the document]
pbabin: I'm asking this to be done at the class level not the interface level, all ConstantExpression implementations and Id class can have this enforcement put in place

pbabin:
Having a check to https://in.talend.com/9082609 there's no indication of the version adding this capability. We should have something like 'since MDM vXXX')

org.talend.mdm.core/src/org/talend/mdm/query/Deserializer.java
```diff
@@ -80,6 +80,10 @@ static ConditionProcessor getProcessor(JsonObject object) {
             return IsProcessor.INSTANCE;
         } else if (object.has("full_text")) { //$NON-NLS-1$
             return FullTextProcessor.INSTANCE;
+        } else if (object.has("in")) { //$NON-NLS-1$
+            return InProcessor.INSTANCE;
+        } else if (object.has("notIn")) { //$NON-NLS-1$
```
pbabin: NotIn shouldn't be an operator by its own as we already have the Not operator

zfkang-talend:
Update the wiki (https://in.talend.com/9082609 ), add the link of jira number.

org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/MappingExpressionTransformer.java
```diff
@@ -288,62 +292,74 @@ public Expression visit(ConstantCollection collection) {
...
+        return getConstant(constant.getValueObject());
     }
 
     @Override
     public Expression visit(DateConstant constant) {
-        return constant;
+        if (constant.isExpressionList()) {
+            return getConstant(constant.getValueList());
+        } else {
+            return getConstant(constant.DATE_FORMAT.format(constant.getValue()));
```
lpw82gf: Should be `DateConstant.DATE_FORMAT` since `DATE_FORMAT` is `public static`.

org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/MappingExpressionTransformer.java
```diff
@@ -288,62 +292,74 @@ public Expression visit(ConstantCollection collection) {
...
+        }
     }
 
     @Override
     public Expression visit(DateTimeConstant constant) {
-        return constant;
+        if (constant.isExpressionList()) {
+            return getConstant(constant.getValueList());
+        } else {
+            return getConstant(constant.DATE_FORMAT.format(constant.getValue()));
```
lpw82gf: Should be `DateTimeConstant.DATE_FORMAT` since `DATE_FORMAT` is `public static`.

org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/MappingExpressionTransformer.java
```diff
@@ -288,62 +292,74 @@ public Expression visit(ConstantCollection collection) {
...
+        return getConstant(constant.getValueObject());
     }
 
     @Override
     public Expression visit(TimeConstant constant) {
-        return constant;
+        if (constant.isExpressionList()) {
+            return getConstant(constant.getValueList());
+        } else {
+            return getConstant(constant.TIME_FORMAT.format(constant.getValue()));
```
lpw82gf: Should use `TimeConstant.TIME_FORMAT`.

org.talend.mdm.core/src/com/amalto/core/query/user/BooleanConstant.java
```diff
@@ -10,23 +10,43 @@
...
+        assert value != null;
+        this.valueList = value;
+        this.value = null;
     }
 
     public Boolean getValue() {
         return value;
     }
 
+    @Override public List<Boolean> getValueList() {
```
lpw82gf: Format!

org.talend.mdm.core/src/com/amalto/core/query/user/ByteConstant.java
```diff
@@ -35,7 +47,15 @@ public boolean cache() {
     }
 
     public Byte getValue() {
-        return constant;
+        return value;
+    }
+
+    @Override public List<Byte> getValueList() {
```
lpw82gf: Format issue as above.

org.talend.mdm.core/src/com/amalto/core/query/user/DateConstant.java
```diff
@@ -26,20 +27,38 @@
...
+        assert valueList != null;
+        this.valueList = valueList;
+        this.value = null;
+    }
+
     public Date getValue() {
         return value;
     }
 
+    @Override public List<Date> getValueList() {
```
lpw82gf: Format issue as above.

org.talend.mdm.core/src/com/amalto/core/query/user/DoubleConstant.java
```diff
@@ -35,7 +47,15 @@ public boolean cache() {
     }
 
     public Double getValue() {
-        return constant;
+        return value;
+    }
+
+    @Override public List<Double> getValueList() {
```
lpw82gf: Format issue as above.

pbabin:
I still don't see any enforcement to forbid illegal calls on values when it's a list and vice versa

PR #385, opened 2017-10-11
TMDM-7196 Can't import a file from Import/Export capability by browser Safari

yjli-talend:
1 Submit button doesn't work,because we can not get file field value when submit form.
2 Waiting bar can not hidden after import file.

**What is the new behavior?**
1 We can get file field value after click file field.Our solution is let file field get focus when mouse over submit button.And then we can get file field value when we submit form.
2 Miss MIME type will cause an error in safari,it will interrupt response in Safari.But it only cause an warning in other browser,so it doesn't work only in safari.

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:


qcui:
LGTM

PR #408, opened 2017-11-14
Revert "TMDM-11397 Change error log to warning log."

sbliu:
Reverts Talend/tmdm-server-se#406


sbliu:
@pbabin As you said it should be a error

PR #415, opened 2017-11-27
TMDM-11022 track/report login activity of the users on the MDM Web UI

zfkang-talend:

**What is the new behavior?**
Add the audit log when login/logout

- [ ] Bugfix
- [x] Feature

- [ ] Yes
- [x] No

**Other information**:


org.talend.mdm.core/src/com/amalto/core/audit/MDMAuditLogger.java
```diff
@@ -0,0 +1,49 @@
...
+import org.talend.logging.audit.AuditLoggerFactory;
+import org.talend.logging.audit.Context;
+import org.talend.logging.audit.ContextBuilder;
+import org.talend.mdm.commmon.util.core.MDMConfiguration;
+
+import com.amalto.core.audit.logger.impl.MDMEventAuditLogger;
+
+public class MDMAuditLogger {
+
+    private static final String SYSPROP_CONFIG_FILE = "talend.logging.audit.config";
```
pbabin: Can it be renamed to something more meaningful?
`AUDIT_CONFIG_FILE_LOCATION`

org.talend.mdm.core/src/com/amalto/core/server/security/jaas/AbstractLoginModule.java
```diff
@@ -188,6 +191,11 @@ public final boolean abort() {
 
     public final boolean logout() {
         subject.getPrincipals().remove(principal);
+        try {
+            MDMAuditLogger.logoutSuccess(LocalUser.getLocalUser().getUsername());
```
pbabin: we have just above the MDMPrincipal object , so there shouldn't be any need to call LocalUser method?

pbabin:
Since we would need to easily disable the functionality, we could rely on the presence of the property or not:
```
    private static final Logger LOGGER = Logger.getLogger(MDMAuditLogger.class);
    private static final String AUDIT_CONFIG_FILE_LOCATION = "talend.logging.audit.config";
    private static final MDMEventAuditLogger auditLogger;
    
    static {
        String auditConfigFileLocation = MDMConfiguration.getConfiguration().getProperty(AUDIT_CONFIG_FILE_LOCATION);
        if (auditConfigFileLocation == null) {
            LOGGER.warn("Audit is disabled.");
            auditLogger = null;
        } else {
            auditConfigFileLocation = SystemPropertyUtils.resolvePlaceholders(auditConfigFileLocation);
            LOGGER.info("Configuring audit using file '" + auditConfigFileLocation + "'");
            System.setProperty(AUDIT_CONFIG_FILE_LOCATION, auditConfigFileLocation);
            auditLogger = AuditLoggerFactory.getEventAuditLogger(MDMEventAuditLogger.class);
        }

        public static void loginSuccess(String userName) {
            if(auditLogger != null) {
              Context ctx = ContextBuilder.create("user", userName).build();
              auditLogger.loginSuccess(ctx);
            }
        }

        ...
    }
```

org.talend.mdm.core/src/com/amalto/core/server/security/jaas/AbstractLoginModule.java
zfkang-talend: I have debug code, when login success, will invoke the mothod commit(), in commit() method, reset() method invoked, principal and username  set to null, in line 193, the variable principal is null, subject.getPrincipals().remove(principal) can't remove it form subject.getPrincipals().

pbabin: Variable `principal`is `null`? Seems very weird

zfkang-talend: Yes,  the login logic is invkoed the javax.security.auth.login.LoginContext.login()(code in the below), first invkoed the login method, next invkoed the commit method, when invkoed the commit method, set the variable to null:
```java
public void login() throws LoginException {

        loginSucceeded = false;

        if (subject == null) {
            subject = new Subject();
        }

        try {
            // module invoked in doPrivileged
            invokePriv(LOGIN_METHOD);
            invokePriv(COMMIT_METHOD);
            loginSucceeded = true;
        } catch (LoginException le) {
            try {
                invokePriv(ABORT_METHOD);
            } catch (LoginException le2) {
                throw le;
            }
            throw le;
        }
    }
```
commit() code: https://github.com/Talend/tmdm-server-se/blob/9bcc6c6b9ca0ced659e41e25c64b1fa881d49921/org.talend.mdm.core/src/com/amalto/core/server/security/jaas/AbstractLoginModule.java#L166 invoked the reset to set variables to null

org.talend.mdm.core/pom.xml
```diff
@@ -318,6 +320,10 @@
             <groupId>org.talend.mdm</groupId>
             <artifactId>org.talend.mdm.commons.core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.talend.daikon</groupId>
+            <artifactId>daikon-audit</artifactId>
```
pbabin: This dep must be in EE
A default impl is to be provided in CE

org.talend.mdm.core/pom.xml
```diff
@@ -87,6 +88,7 @@
                                 <include>com/amalto/core/delegator/impl/DefaultLocalUserDelegator.class</include>
                                 <include>com/amalto/core/save/generator/AutoIncrementGenerator.class</include>
                                 <include>com/amalto/core/server/ServerAccessImpl*.class</include>
+                                <exclude>com/amalto/core/audit/MDMAuditLogger.class</exclude>
```
pbabin: The interface shouldn't be exclude here

org.talend.mdm.core/pom.xml
```diff
@@ -73,6 +73,7 @@
                                 <exclude>com/amalto/core/delegator/impl/DefaultLocalUserDelegator.class</exclude>
                                 <exclude>com/amalto/core/save/generator/AutoIncrementGenerator.class</exclude>
                                 <exclude>com/amalto/core/server/ServerAccessImpl*.class</exclude>
+                                <exclude>com/amalto/core/audit/MDMAuditLogger.class</exclude>
```
pbabin: The interface shouldn't be exclude here

zfkang-talend: The implemention type is like AutoIncrementGenerator.java, in CE and EE have one same name file, but for default-jar, exclude it, only compile in open-jar.

PR #490, opened 2018-03-07
TMDM-11813 Super Admin has no access to Custom Layout linked to a View

zfkang-talend:
Jira: https://jira.talendforge.org/browse/TMDM-11813

Can not get the custom layout for super admin user.

**What is the new behavior?**
For super admin, set all custom layout to contain permissions when fetched.

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/src/com/amalto/core/server/DefaultCustomForm.java
+++ b/org.talend.mdm.core/src/com/amalto/core/server/DefaultCustomForm.java
@@ -17,6 +17,7 @@ import java.util.Set;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
+import org.talend.mdm.commmon.util.core.ICoreConstants;
 
 import com.amalto.core.delegator.ILocalUser;
 import com.amalto.core.objects.ObjectPOJO;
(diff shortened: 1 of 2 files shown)


org.talend.mdm.core/src/com/amalto/core/server/DefaultCustomForm.java
```diff
@@ -91,21 +92,28 @@ public CustomFormPOJO getUserCustomForm(CustomFormPOJOPK cpk) throws XtentisExce
...
-                //get Specifications for the View Object
-                RoleSpecification specification = role.getRoleSpecifications().get(objectType);
-                if (specification != null) {
-                    if (specification.isAdmin()) {
-                        allCustomFormPermitted = true;
+
+            boolean isSuperAdmin = roleNames.contains(ICoreConstants.ADMIN_PERMISSION)
+                    && roleNames.contains(ICoreConstants.SYSTEM_ADMIN_ROLE);
+            if (isSuperAdmin) {
+                allCustomFormPermitted = true;
```
pbabin: what about users with 'administration' role only?

lpw82gf: Since
- We can't create user with "administration" role only on TAC.
- From migration also can't have that kind user.
So do we need to consider that case?

zfkang-talend:
fetch customer layout is invoked this method,  https://github.com/Talend/tmdm-server-se/blob/68d967dc9a0a183c790d63b3543437473106d462/org.talend.mdm.webapp.browserecords/src/main/java/org/talend/mdm/webapp/browserecords/server/actions/BrowseRecordsAction.java#L1815 if the parameter of customFormName is empty, will return null, so in CE editor, have no customer layout, this parameter is always empty.

PR #640, opened 2018-09-21
TMDM-12572 Wrong value returned by MDM REST API select query when 2 fields with same name in 2 different entities.

htyin:
https://jira.talendforge.org/browse/TMDM-12572

**What is the new behavior?**

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core.storage.sql/src/test/java/com/amalto/core/query/StorageQueryTest.java
+++ b/org.talend.mdm.core.storage.sql/src/test/java/com/amalto/core/query/StorageQueryTest.java
@@ -5633,6 +5633,52 @@ public class StorageQueryTest extends StorageTestCase {
         }
     }
 
+    // TMDM-12572 Wrong value returned by MDM REST API select query when 2 fields with same name in 2 different
+    // entities.
+    public void testQueryFKFieldWithSameName() throws Exception {
+        MetadataRepository repository = new MetadataRepository();
+        repository.load(DataRecordCreationTest.class.getResourceAsStream("StorageQueryTest_3.xsd"));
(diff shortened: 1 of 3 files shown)


org.talend.mdm.core.storage.sql/src/test/java/com/amalto/core/query/StorageQueryTest.java
```diff
@@ -5633,6 +5633,52 @@ public void testJoinQueryWithForeignKey() throws Exception {
...
+                .join(principalSalesAgencyMapping.getField("bfoeSalesAgencyUuid"), salesAgency.getField("uuid"));
+
+        StorageResults results = storage.fetch(qb.getSelect());
+        assertEquals(1, results.getCount());
+        for (DataRecord record : results) {
+            assertEquals("e37e995a-1389-4ad9-a37b-9a4b87f8bbe1", record.get("PrincipalLegalEntityUuid"));
+            assertEquals("26160961-e4a7-438b-8c92-cae8340d8664", record.get("SalesAgencyLegalEntityUuid"));
+        }
+    }
+
```
hw-zhu: For this case, we need to check 1) whether correct to generate the index. 2) There is impact for fuzzy search.

htyin: The issue is caused by using field name as key of alias map. When there are fields with the same name in different entity, error happened in getting alias. Now the key has been changed to entityName/fieldName. It should have nothing to do with index.

PR #657, opened 2018-10-09
TMDM-12624 Issue with date field defined as a primary key

VictorLi-talendbj:

If we used date, dateTime or time type as primary key in Studio data model, then from MDM Web UI, we cannot create or view the record. There will have exception reported from MDM server.

https://jira.talendforge.org/browse/TMDM-12624

**What is the new behavior?**

The fixing will allow date, dateTime and time as primary key, the record can be created or viewed from Web UI.

It added a conversion from Java.util.Date to Java.sql.Timestamp as required.

- [X] Bugfix
- [ ] Feature

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/src/com/amalto/core/storage/StorageMetadataUtils.java
+++ b/org.talend.mdm.core/src/com/amalto/core/storage/StorageMetadataUtils.java
@@ -44,6 +44,7 @@ import org.talend.mdm.commmon.metadata.SimpleTypeFieldMetadata;
 import org.talend.mdm.commmon.metadata.TypeMetadata;
 import org.talend.mdm.commmon.metadata.Types;
 
+import com.amalto.core.query.user.ConstantExpression;
 import com.amalto.core.query.user.DateConstant;
 import com.amalto.core.query.user.DateTimeConstant;
 import com.amalto.core.query.user.TimeConstant;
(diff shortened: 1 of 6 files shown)


org.talend.mdm.core/src/com/amalto/core/storage/StorageMetadataUtils.java
```diff
@@ -664,6 +665,19 @@ public static Object convert(String dataAsString, String type) {
...
+                public java.lang.Object transform(java.lang.Object input) {
+                        return new Timestamp(((Date)input).getTime());
+                    }
+                });
+            return constant.getValueList();
+        } else {
+            return new Timestamp(constant.getValue().getTime());
+        }
+    }
+
```
hw-zhu: public static <T> T convert(ConstantExpression<Date> constant) {
        if (constant.isExpressionList()) {
                     ........
            return (T)constant.getValueList();
        } else {
            return (T)new Timestamp(constant.getValue().getTime());
        }    
    }

caller:
Timestamp d = StorageMetadataUtils.<Timestamp>convert(xxxx);
List<Timestamp> dd = StorageMetadataUtils.<List<Timestamp>>convert(xxxx);

Try to implement it with static generic method.

org.talend.mdm.core/test/com/amalto/core/storage/StorageMetadataUtilsTest.java
```diff
@@ -146,4 +157,64 @@ public void testGetIds() {
...
+    @Test
+    public void testConvertConstantExpressionDate() throws ParseException {
+        DateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd"); //$NON-NLS-1$
+
+        DateConstant dateConstant = new DateConstant("2018-10-06"); //$NON-NLS-1$
+        Object dateStamp = StorageMetadataUtils.convert(dateConstant);
+        assertTrue(dateStamp instanceof Timestamp);
+
+        List<Date> listDate = new LinkedList<Date>();
+        listDate.add(DATE_FORMAT.parse("2018-10-11")); //$NON-NLS-1$
```
lpw82gf: We may don't need three values in the list, one would be OK.

VictorLi-talendbj: Done.

PR #662, opened 2018-10-16
TMDM-12666 Error when trying to open a journal record after deletion of an occurence of [0..many]

hw-zhu:
…of an occurence of [0..many]

There is NPR when entity has the same property name.

**What is the new behavior?**
Fixed above issue.
The root cause is that there are the same property name in entity hierarchy， so the exception arise when the W3C Document Object Model parse the XML schemal.

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/test/com/amalto/core/history/accessor/ManyFieldAccessorTest.java
+++ b/org.talend.mdm.core/test/com/amalto/core/history/accessor/ManyFieldAccessorTest.java
@@ -20,25 +20,30 @@ import com.amalto.core.save.DOMDocument;
 import com.amalto.core.util.Util;
 
 public class ManyFieldAccessorTest {
-    
+
     @Test
     public void testCreateEmptyChildNode() throws Exception {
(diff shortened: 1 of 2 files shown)


org.talend.mdm.core/test/com/amalto/core/history/accessor/ManyFieldAccessorTest.java
```diff
@@ -27,18 +27,33 @@ public void testCreateEmptyChildNode() throws Exception {
...
         Accessor accessor = document.createAccessor("/Communications/TypeCommunication[1]"); //$NON-NLS-1$
         accessor.create();
-        
+
         Assert.assertNotNull(document);
     }
-    
+
     private DOMDocument getTypeWithNoElementContent() throws IOException, SAXException, ParserConfigurationException {
         String xml =
```
lpw82gf: Refactor the xml string to be inline.

PR #722, opened 2018-12-10
TMDM-12877 [Impact analysis] Modify Product/Features from 1-1 to 0-1 redeploy DM failed.

zfkang-talend:
Jira: https://jira.talendforge.org/browse/TMDM-12877

Modify **ContainedTypeFieldMetadata** type field, which like **Features** field or **Stores** field in Product data model, have 2 issues:

1. Failed to move this field from 1-1 to 0-1:
    Reason: sometimes liquibase to execute changeLog, but this field is not existed.
2. Failed to move this field from 0-1 to 1-1:
    Reason: failed to fetch count existed in database of this field, because it is not existed.
3. Failed to move Product/Feature/Colors/Color or Product/Feature/Size/Size from 1-1 to 0-1.
     Reason: liquibase get the wrong table name and column name, so it's failed when liquibase execute.

**Why Not Existed:**
Because it is a **ContainedTypeFieldMetadata** type, it depends on the DM and database, for example Product data model:
In Product Table, have no column for **Features** , but have two table: **product_x_features_colors_color** and **product_x_features_sizes_size**, these two table's key is product table's key: like below:
```

mysql> select * from product;
+------+-----------+--------+---------------+----------------+---------+---------------+---------------+--------------------+------------------+
| x_id | x_picture | x_name | x_description | x_availability | x_price | x_family_x_id | x_onlinestore | x_talend_timestamp | x_talend_task_id |
+------+-----------+--------+---------------+----------------+---------+---------------+---------------+--------------------+------------------+
| 1    | NULL      | name   | description   | NULL           |    1.20 | NULL          | NULL          |      1544434182360 | NULL             |
| 20   | NULL      | name   | description   | NULL           |    1.20 | NULL          | NULL          |      1544434294410 | NULL             |
+------+-----------+--------+---------------+----------------+---------+---------------+---------------+--------------------+------------------+
2 rows in set (0.00 sec)

mysql> select * from product_x_features_colors_color;
+------+------------+-----+
| x_id | value      | pos |
+------+------------+-----+
| 1    | Light Blue |   0 |
| 20   | Light Blue |   0 |
+------+------------+-----+
2 rows in set (0.00 sec)

mysql> select * from product_x_features_sizes_size;
+------+-------+-----+
| x_id | value | pos |
+------+-------+-----+
| 1    | Small |   0 |
| 20   | Small |   0 |
+------+-------+-----+
2 rows in set (0.00 sec)
```
However, for ORACLE database, it existed the column '_X_FEATURES_X_TALEND'_.
Have other case, the column is existed in main table.

**What is the new behavior?**
For above two issues, the solution are below:
No.1: 
    Can not add the Change into the ChangeSet(which used Liquibase), use Hibernate tools to update  if the column existed (oracle contains the columns 'X_FEATURES_X_TALEND_ID' for feature)
No.2:
   Avoid to fetch the number existed in database, because when moved 0-1 to 1-1, it must high, whatever existed data.
No.3:
   Modified to use hibernate tool to updated the not-null properties

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:


org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/HibernateStorageDataAnaylzer.java
```diff
@@ -107,9 +107,12 @@ private int fetchFieldCountOfNull(ComplexTypeMetadata entry, FieldMetadata field
     private int fetchFieldCountByUserQuery(UserQueryBuilder qb) {
         int count = 0;
         storage.begin();
-        StorageResults results = storage.fetch(qb.getSelect());
+        StorageResults results = null;
         try {
+            results = storage.fetch(qb.getSelect());
             count = results.getCount();
+        } catch (Exception e) {
+            LOGGER.error("Fetch field's count failed.", e); //$NON-NLS-1$
```
VictorLi-talendbj: Failed to fetch count of field.

PR #731, opened 2019-01-02
TMDM-12935 [REST API] Support JSON on batch Create/Update records

VictorLi-talendbj:

This is a new feature, see
https://jira.talendforge.org/browse/TMDM-12935

The resource "POST /data/{containerName}/batch" can only support XML format input, such as:
```
<results>
    <Product>...</Product>
    <Product>...</Product>
    <Store>...</Store>
</results>
```
**What is the new behavior?**

With this ticket, we added JSON format for the resource:
POST /data/{containerName}/batch
So this will comply with others APIs of create/update, the format JSON would be supported, such as:
```
[
	{
		"Product": {
			"Id": "1011",
			"Name": "Talend Dog T-Shirt",
			"Price": "16.99",
			"Family": "[3]",
			"Stores": {
				"Store": [
					"[TalendUS]",
					"[TalendFR]"
				]
			}
		}
	},
	{
		"Product": {
			"Id": "1024",
			"Name": "Talend Dog T-Shirt",
			"Availability": "false",
			"Price": "16.99",
			"Family": "[3]",
		}
	},
	{
		"Store": {
			"Id": "789-store",
			"Address": "new-store"
		}
	}	
]
```

- [ ] Bugfix
- [X] Feature

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/src/com/amalto/core/storage/record/DataRecordJSONReader.java
+++ b/org.talend.mdm.core/src/com/amalto/core/storage/record/DataRecordJSONReader.java
@@ -22,7 +22,9 @@ import org.apache.log4j.Logger;
 import org.talend.mdm.commmon.metadata.*;
 
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Iterator;
+import java.util.Set;
 import java.util.Map.Entry;
 
 import com.google.gson.JsonArray;
@@ -40,6 +42,8 @@ public class DataRecordJSONReader implements DataRecordReader<JsonElement> {
 
     private String entityName = null;
 
+    private static Set<String> tagNames = null;
+
     public DataRecordJSONReader() {
     }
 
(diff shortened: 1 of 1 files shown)


org.talend.mdm.core/src/com/amalto/core/storage/record/DataRecordJSONReader.java
```diff
@@ -178,24 +184,21 @@ private void readJsonPrimitive(MetadataRepository repository, DataRecord dataRec
...
-        if (!textContent.isEmpty()) {
-            if (type.hasField(tagName)) {
-                FieldMetadata field = type.getField(tagName);
-                if (field instanceof ReferenceFieldMetadata) {
-                    ComplexTypeMetadata actualType = ((ReferenceFieldMetadata) field).getReferencedType();
-                    dataRecord.set(field, StorageMetadataUtils.convert(textContent, field, actualType));
-                } else {
-                    dataRecord.set(field, StorageMetadataUtils.convert(textContent, field, type));
-                }
+        if (type.hasField(tagName)) {
```
hw-zhu: Whether the outer the xml tag may embed the same name sub tag? like 

> <product><name></name><feature><name></name></feature></product>

VictorLi-talendbj: The XML tag is possible to have the format as `<product><name></name><feature><name></name></feature></product>`, if so, we also need to set value for the `<name>` which wouldn't be changed when update a record. For example, we can have input as: `<product><name>original_name</name><feature><name></name></feature></product>`, so the result of update is that the name of product won't change, but the name of feature will be removed.

PR #749, opened 2019-02-12
TMDM-13124 [RestAPI] GET /data/{containerName}/{type}/{id} fails with HTTP 500

zfkang-talend:
Jira: https://jira.talendforge.org/browse/TMDM-13124

When use history storage fetches records, our logic step:
1. get the data record status with the fetch date.
2. create  temporary storage
3. fill data into temporary  --> **failed** to fill record into temporary)
4. fetch data.

Reason:
1. Update the record to empty for Reference Field
```
 Service
        |__Service ID(Simple Field)
        |__Address(Reference type)
              |__Address ID
              |__Address Name
```
    
At first, Use the Service ID get the data record of Service:
```          
          field=Service ID, Value=1
          field=Address, Value=[1]
          field=Address ID,value=1
          field=Address Name, value=Beijing
```   
Update the Service Data record to:
```       
          field=Service ID, Value=1
          field=Address, Value=[1]
          field=Address ID,value=1
```
other field of **Address** (such as field **Address Name**) will be removed

2. get the wrong child node
```
      Organisation
           |__Address(Reference)
           |__Service(Reference)
                  |__Address(Reference)
```
System used Map to store child node, first parse entity _Organisation_， put _Address_ into map, but parse entity _Service_, Address existed, don't add Address into Service's child node.

3. wrong result about method **hasPrivious()** when judge UserAction with update report:
```
    action[0]=CreateAction
    action[1]=NoOpAction
```
create record, have above two actions, only  if the current action index=0, invoke hasPrivious(), will return true. it's wrong. 

4. If it's the self-reference, endless loop 

**What is the new behavior?**
for above 4 problems, the solution:
1. if the reference field's data contains, and the key field value has no change, don't update.
2. if the child node had been parsed, copy it and added to child node. (this modify in ee)
3. add the condition if actions length == 2. (this modify in ee)
4. remove the child node which is empty data record. (this modify in ee)

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:


org.talend.mdm.core/src/com/amalto/core/history/accessor/record/ManyValue.java
```diff
@@ -39,11 +39,20 @@ public void set(MetadataRepository repository, DataRecord record, PathElement el
...
         if (element.field instanceof ReferenceFieldMetadata) {
-            ComplexTypeMetadata referencedType = ((ReferenceFieldMetadata) element.field).getReferencedType();
-            DataRecord referencedRecord = (DataRecord) StorageMetadataUtils.convert(String.valueOf(value), element.field, referencedType);
-            list.set(element.index, referencedRecord);
+            ReferenceFieldMetadata fieldMetadata = (ReferenceFieldMetadata) element.field;
+            DataRecord oldRecord = (DataRecord) list.get(element.index);
+            boolean needResetValue = true;
+            if (oldRecord != null) {
+                String oldValue = String.valueOf(oldRecord.get(fieldMetadata.getReferencedField()));
+                needResetValue = !value.equals('[' + oldValue + ']');
```
VictorLi-talendbj: Would suggest to not use additional boolean, just have another `if` condition, and use StringUtils.equals to check string values.

org.talend.mdm.core/src/com/amalto/core/history/accessor/record/ManyValue.java
```diff
@@ -39,11 +39,20 @@ public void set(MetadataRepository repository, DataRecord record, PathElement el
...
+            }
+            if (needResetValue) {
+                ComplexTypeMetadata referencedType = fieldMetadata.getReferencedType();
+                DataRecord referencedRecord = (DataRecord) StorageMetadataUtils.convert(value, element.field, referencedType);
+                list.set(element.index, referencedRecord);
+            }
         } else {
-            list.set(element.index, StorageMetadataUtils.convert(String.valueOf(value), element.field));
+            list.set(element.index, StorageMetadataUtils.convert(value, element.field));
```
VictorLi-talendbj: Not sure why removed `Strinng.valueOf` here, but not in another file `SimpleValue.java`

lpw82gf: `value` is String already.

org.talend.mdm.core/src/com/amalto/core/history/accessor/record/SimpleValue.java
```diff
@@ -28,9 +28,17 @@ public void set(MetadataRepository repository, DataRecord record, PathElement el
...
+                String oldValue = String.valueOf(record.get(fieldMetadata.getReferencedField()));
+                needResetValue = !value.equals('[' + oldValue + ']');
+            }
+            if (needResetValue) {
+                ComplexTypeMetadata referencedType = fieldMetadata.getReferencedType();
+                DataRecord referencedRecord = (DataRecord) StorageMetadataUtils.convert(value, element.field, referencedType);
+                record.set(element.field, referencedRecord);
+            }
         } else {
             record.set(element.field, StorageMetadataUtils.convert(String.valueOf(value), element.field));
```
lpw82gf: Should also remove `String.valueOf` here.

org.talend.mdm.core/src/com/amalto/core/history/accessor/record/ManyValue.java
lpw82gf: `value` is sure can't be `null`, so `equals` would be enough here.

PR #771, opened 2019-03-19
TMDM-13289 [REST API] APIs of Staging area management doesn't work

yjli-talend:
https://jira.talendforge.org/browse/TMDM-13289
Get user by id,but we can't get id when execute rest api. we can only get username from `LocalUser.getLocalUser().getIdentity()`

**What is the new behavior?**
Use existed user xml to get cluster and model information. We have a auto test to cover this case.

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:


org.talend.mdm.core/src/com/amalto/core/util/Util.java
```diff
@@ -762,18 +762,12 @@ public static String getBarHomeDir() {
         return ret;
     }
 
-    private static Element getLoginProvisioningFromDB() throws Exception {
```
VictorLi-talendbj: How about still keep this function, but rename it as `getLoginProvisioningFromXML`? And do we need to check if `LocalUser.getLocalUser().getUserXML()` will be null?

PR #784, opened 2019-04-04
TMDM-13365 Error while deleting a record

hw-zhu:
https://jira.talendforge.org/browse/TMDM-13365
There is a error when deleting a record of a special Data Model, The key point is there is a NPE when create query condition. 

**What is the new behavior?**
The operation may works normally for now, added a a snippet of code to check if there is null value in the condition, if it exists, don't add the item to the query condition. 

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/StandardQueryHandler.java
+++ b/org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/StandardQueryHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2006-2018 Talend Inc. - www.talend.com
+ * Copyright (C) 2006-2019 Talend Inc. - www.talend.com
  * 
  * This source code is available under agreement available at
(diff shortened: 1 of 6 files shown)


org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/StandardQueryHandler.java
```diff
@@ -904,7 +904,9 @@ public StorageResults visit(Isa isa) {
     @Override
     public StorageResults visit(Compare condition) {
         Criterion criterion = condition.accept(criterionVisitor);
-        criteria.add(criterion);
+        if (criterion != null) {
```
VictorLi-talendbj: Can this be reproduced in FullTextQueryHandler?

hw-zhu: After testing, Lucene don't use the same logic snippet of code.

PR #843, opened 2019-07-10
TMDM-13587 Can't save FK under the reusable type with customer's data model

hw-zhu:
https://jira.talendforge.org/browse/TMDM-13587
What is the current behavior?** (You should also link to an open issue here)
Can't save the FK under the reusable type, The key point is the column FK don't be mapped correctly when creating a Java Class object in memory based on the Hibernate mapping file.

**What is the new behavior?**
Modified the snippet of code to generate the correct column name, then can save the FK under the reusable type.

- [*] The commit message follows Talend standard
- [*] Tests for the changes have been added (for bug fixes / features)

- [*] Bugfix
- [ ] Feature

- [ ] Yes
- [*] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/StatefulContext.java
+++ b/org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/StatefulContext.java
@@ -10,13 +10,16 @@
 
 package com.amalto.core.storage.hibernate;
 
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 import org.talend.mdm.commmon.metadata.ComplexTypeMetadata;
+import org.talend.mdm.commmon.metadata.ContainedComplexTypeMetadata;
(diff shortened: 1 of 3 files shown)


org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/StatefulContext.java
```diff
@@ -66,6 +71,42 @@ public String getFieldColumn(FieldMetadata field) {
...
+     * @return: the actual key info
+     */
+    private static String getQualifiedKey(FieldMetadata field) {
+        String key = field.toString();
+        ComplexTypeMetadata type = field.getContainingType();
+        if (type == null) {
+            return key;
+        }
+        ComplexTypeMetadata topLevelType = (ComplexTypeMetadata) MetadataUtils.getSuperConcreteType(type);
+        boolean isDupFKName = false;
```
VictorLi-talendbj: What's the meaning of `isDupFKName`? can have an example or explanation please?

hw-zhu: It apply to a special DM, eg:  a complex type BaseType including a field that is a FK type, meanwhile, the complex type BaseType be inherited by multiple sub complex type, like FirstType, SecondType, so this two sub complex type refer to the same FK field. that's the usage of variable isDupFKName.

PR #851, opened 2019-08-02
TMDM-12997 Add more fields in the Jounal from the MDM Web UI

lpw82gf:
- No primary key info on Journal
- Can't search Journal by UserName

**What is the new behavior?**
- Added primary key info to Journal
- Could search Journal by UserName

- [ ] Bugfix
- [x] Feature

- [ ] Yes
- [x] No

**Other information**:


org.talend.mdm.webapp.journal/src/main/java/org/talend/mdm/webapp/journal/server/service/JournalDBService.java
```diff
@@ -80,11 +86,22 @@ public JournalDBService(WebService webService) {
...
-            sortDir = Constants.SEARCH_DIRECTION_ASC;
-        } else if (SortDir.DESC.equals(SortDir.findDir(sort))) {
-            sortDir = Constants.SEARCH_DIRECTION_DESC;
+
+        // Ignore Primary Key Info if using Oracle DB, CLOB doesn't support ORDER BY
+        if ("primaryKeyInfo".equals(field)) {
+            StorageAdmin storageAdmin = ServerContext.INSTANCE.get().getStorageAdmin();
+            RDBMSDataSource dataSource = (RDBMSDataSource) storageAdmin.get(UpdateReportPOJO.DATA_CLUSTER, StorageType.MASTER).getDataSource();
+            if (HibernateStorageUtils.isOracle(dataSource.getDialectName())) {
+                field = null;
```
VictorLi-talendbj: If Oracle doens't support ORDER BY, do we need to continue? or can we just return now?

lpw82gf: The action is in progress already, if we just return now, there is no data returned then.

lpw82gf: If we want to disable that, it is better to stop from front-end, but in front-end we can't get what DB used, but pass the DB to front-end isn't a good choice.
As we will document that, Oracle doesn't support ORDER BY, so don't think it would be a big problem here.

PR #886, opened 2019-09-23
TMDM-14011 Error Unauthorized write access by user xxx Item '__SYSTEM.User.ac36a53c-5c26-4afa-a9eb-f5c478a2ed04' when Admin Permissions On all Instances not checked for Object Type=item

lpw82gf:
https://jira.talendforge.org/browse/TMDM-14011
- Uncheck  `Admin Permissions On all Instances` for `Item` current user can't save its own information like: Language and Portal Configuration

**What is the new behavior?**
- Read Only web user should be able to save its own information

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/src/com/amalto/core/objects/ItemPOJO.java
+++ b/org.talend.mdm.core/src/com/amalto/core/objects/ItemPOJO.java
@@ -665,27 +665,31 @@ public class ItemPOJO implements Serializable {
         assert user != null;
         boolean authorizedAccess;
         String username = user.getUsername();
+        String dataCluster = itemPOJOPK.getDataClusterPOJOPK().getUniqueId();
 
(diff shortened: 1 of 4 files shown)


org.talend.mdm.core/src/com/amalto/core/objects/ItemPOJO.java
```diff
@@ -665,27 +666,32 @@ private static void checkAccess(ILocalUser user, ItemPOJOPK itemPOJOPK, boolean
         assert user != null;
         boolean authorizedAccess;
         String username = user.getUsername();
+        String dataCluster = itemPOJOPK.getDataClusterPOJOPK().getUniqueId();
```
VictorLi-talendbj: Do we need to have condition check `itemPOJOPK.getDataClusterPOJOPK() != null` as before?

lpw82gf: Don't think so, and for the old code, that check is useless too.

PR #893, opened 2019-09-25
TMDM-13991 Advanced Search does not work with AND query on Foreign Key Field

hw-zhu:
https://jira.talendforge.org/browse/TMDM-13991
Lucene query need to use full path as condition based on above description,but now only have the Node and the element below the Node when the current FieldMetadata element to be parsed

**What is the new behavior?**
Modify related method, get the current field's containingType, if it's not empty, continue to rise up until it's containingType is empty, finally, get the absolute path of the field, get the absolute path for the current fieldName.

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/LuceneQueryGenerator.java
+++ b/org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/LuceneQueryGenerator.java
@@ -415,6 +415,38 @@ class LuceneQueryGenerator extends VisitorAdapter<Query> {
         return query;
     }
 
+    /**
+     * Get a full path of field for its directory tree.
+     * <pre>
+     * Person
+     *      |__Stores
+     *              |__Store
+     *                      |__Name
+     *      |__Age                
+     * </pre>
+     * @return: Person.Stores.Store.Name
+     */
+    private String getFullPathName(FieldFullText fieldFullText) {
+        FieldMetadata fieldMetadata = fieldFullText.getField().getFieldMetadata();
+        StringBuilder fullPathFieldName = new StringBuilder();
+        while (fieldMetadata instanceof ReferenceFieldMetadata) {
+            String fieldName = fieldMetadata.getName();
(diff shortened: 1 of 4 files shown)


org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/LuceneQueryGenerator.java
```diff
@@ -415,6 +415,38 @@ public Void visit(EnumerationFieldMetadata enumField) {
...
+     * Person
+     *      |__Stores
+     *              |__Store
+     *                      |__Name
+     *      |__Age                
+     * </pre>
+     * @return: Person.Stores.Store.Name
+     */
+    private String getFullPathName(FieldFullText fieldFullText) {
+        FieldMetadata tempFieldMetadata = fieldFullText.getField().getFieldMetadata();
```
VictorLi-talendbj: How about use variable name `filedMetadata`?

org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/LuceneQueryGenerator.java
```diff
@@ -415,6 +415,38 @@ public Void visit(EnumerationFieldMetadata enumField) {
...
+        StringBuilder fullPathFieldName = new StringBuilder();
+        while (tempFieldMetadata instanceof ReferenceFieldMetadata) {
+            String tempFieldName = tempFieldMetadata.getName();
```
VictorLi-talendbj: How about use variable as `fieldName`?

hw-zhu: done

org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/LuceneQueryGenerator.java
hw-zhu: done

PR #978, opened 2020-01-14
TMDM-14289 tMDMRollback somtimes let transactions "opened"

VictorLi-talendbj:
https://jira.talendforge.org/browse/TMDM-14289

A job that fails (exception) and calls tMDMRollback sometimes let a transaction "opened" (not commited , not rollbacked). It was found that storage begin may not completed with storage commit or storage rollback in a request if there are many requests from a loop Studio Job.

**What is the new behavior?**

Added thread sleep between each for the transaction in SaverSession, this will ensure that a request can complete the whole process of transaction.

- [X] Bugfix
- [ ] Feature

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/src/com/amalto/core/save/SaverSession.java
+++ b/org.talend.mdm.core/src/com/amalto/core/save/SaverSession.java
@@ -16,6 +16,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import org.apache.log4j.Logger;
 
 import org.talend.mdm.commmon.metadata.FieldMetadata;
 import org.talend.mdm.commmon.util.core.EUUIDCustomType;
--- a/org.talend.mdm.core/src/com/amalto/core/save/SaverSession.java
+++ b/org.talend.mdm.core/src/com/amalto/core/save/SaverSession.java
@@ -62,10 +62,24 @@ public class SaverSession {
 
     private boolean hasMetAutoIncrement = false;
 
-    private static final long TRANSACTION_WAIT_MILLISECONDS;
+    private static final String TRANSACTION_WAIT_MILLISECONDS_CONFIG;
 
     static {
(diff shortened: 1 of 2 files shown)


org.talend.mdm.core/src/com/amalto/core/save/SaverSession.java
```diff
@@ -58,6 +62,12 @@
 
     private boolean hasMetAutoIncrement = false;
 
+    private static final long TRANSACTION_WAIT_MILLISECONDS;
+
+    static {
+        TRANSACTION_WAIT_MILLISECONDS = Long.valueOf(MDMConfiguration.getTransactionWaitMilliseconds());
```
pbabin: Signature of the method is `public static long getTransactionWaitMilliseconds()`
no need to re-execute the `Long.ValueOf`
As said in https://github.com/Talend/tmdm-common/pull/162#pullrequestreview-342956932 the method could have been placed here instead of the common code

VictorLi-talendbj: Both changed, please take a look.

org.talend.mdm.core/src/com/amalto/core/save/SaverSession.java
```diff
@@ -215,6 +242,11 @@ public void end(Committer committer) {
                     }
                     throw e;
                 }
+                try {
+                    Thread.sleep(transactionWaitMilliseconds);
```
lpw82gf: Can we `Thread.sleep(getTransactionWaitMilliseconds());` and remove `long transactionWaitMilliseconds = getTransactionWaitMilliseconds();` ?

VictorLi-talendbj: This is to get the value out of the for loop.

lpw82gf: OK

PR #1002, opened 2020-02-03
TMDM-14289 tMDMRollback somtimes let transactions "opened"

VictorLi-talendbj:
https://jira.talendforge.org/browse/TMDM-14289

A job that fails (exception) and calls tMDMRollback sometimes let a transaction "opened" (not commited , not rollbacked). It was found that storage begin may not completed with storage commit or storage rollback in a request if there are many requests from a loop Studio Job.

**What is the new behavior?**

Added transaction process limitation, this will ensure that a request can complete the whole process of transaction.

- [X] Bugfix
- [ ] Feature

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/src/com/amalto/core/util/Util.java
+++ b/org.talend.mdm.core/src/com/amalto/core/util/Util.java
@@ -27,6 +27,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.MissingResourceException;
 import java.util.Properties;
 import java.util.regex.Matcher;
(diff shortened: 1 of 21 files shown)


org.talend.mdm.core/src/com/amalto/core/util/Util.java
```diff
@@ -170,6 +171,26 @@
 
     private static com.amalto.core.server.api.Transformer defaultTransformer;
 
+    private static final AtomicInteger TRANSACTION_CURRENT_REQUESTS = new AtomicInteger();
+
+    private static final String TRANSACTION_WAIT_MILLISECONDS_KEY = "transaction.concurrent.wait.milliseconds"; //$NON-NLS-1$
+
+    public static long TRANSACTION_WAIT_MILLISECONDS_VALUE = 0L;
```
hw-zhu: we may change the parameter to final type to ensure the value cannot be overridden.

org.talend.mdm.core/src/com/amalto/core/util/Util.java
```diff
@@ -1289,4 +1310,27 @@ public static String removeBracketWithNumber(String path) {
...
+                }
+            }
+        }
+    }
+
+    public static void endTransactionLimit() {
+        if (TRANSACTION_WAIT_MILLISECONDS_VALUE > 0) {
+            TRANSACTION_CURRENT_REQUESTS.decrementAndGet();
+        }
+    }
```
hw-zhu: Pls add some comments to ensure the two method cannot be repeated nested invoked in the same thread. otherwise, the second calling will be blocked.

org.talend.mdm.core/src/com/amalto/core/util/Util.java
```diff
@@ -1289,4 +1310,27 @@ public static String removeBracketWithNumber(String path) {
...
+            synchronized (Util.class) {
+                try {
+                    while (TRANSACTION_CURRENT_REQUESTS.get() >= 1) {
+                        Thread.sleep(TRANSACTION_WAIT_MILLISECONDS_VALUE);
+                    }
+                    TRANSACTION_CURRENT_REQUESTS.incrementAndGet();
+                } catch (InterruptedException e) {
+                    if (LOGGER.isDebugEnabled()) {
+                        LOGGER.debug("Failed to sleep thread.", e);
+                    }
```
hw-zhu: If Invoking Thread.sleep() to cause an interruptedException, the interrupted flag also will be clean. In order to sure the following steps, we may add snippet of code `Thread.currentThread().interrupt()` to remark the flag.

org.talend.mdm.core/src/com/amalto/core/util/Util.java
VictorLi-talendbj: If changed to final, then it can't be assigned value in the below static block. So we cannot change it.

org.talend.mdm.core/src/com/amalto/core/util/Util.java
VictorLi-talendbj: Done.

org.talend.mdm.core/src/com/amalto/core/util/Util.java
VictorLi-talendbj: I don't think this is needed in this case as no other function will check the flag.

PR #1038, opened 2020-03-13
TMDM-14507 SOAP/REST API work unexpected when provide value for non-PK autoincrement fields

VictorLi-talendbj:
https://jira.talendforge.org/browse/TMDM-14507

For Entity with Non-PK Auto Increment Fields, when create or update a record via SOAP/REST API, if provided value for the Auto Increment Fields, the provided value would be used, but the System's Auto Increment Value in CONF for that field will be increased which is not same as Auto Increment PK Fields.

**What is the new behavior?**

For Entity with Non-PK Auto Increment Fields, when create or update a record via SOAP/REST API, if provided value for the Auto Increment Fields, the provided value would be used, and the System's Auto Increment Value in CONF for that field would not be increased which is same as Auto Increment PK Fields.

- [X] Bugfix
- [ ] Feature

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core.storage.sql/src/test/java/com/amalto/core/save/DocumentSaveTest.java
+++ b/org.talend.mdm.core.storage.sql/src/test/java/com/amalto/core/save/DocumentSaveTest.java
@@ -671,6 +671,9 @@ public class DocumentSaveTest extends TestCase {
         assertEquals("1", evaluate(committedElement, "/Person/complex/autoIncrement2"));
     }
 
+    /**
+     * TMDM-14507: SOAP/REST API work unexpected when provide value for non-PK autoincrement fields
+     */
     public void testNormalFieldAutoIncrementValueProvided() throws Exception {
         final MetadataRepository repository = new MetadataRepository();
         repository.load(DocumentSaveTest.class.getResourceAsStream("metadata24.xsd"));
--- a/org.talend.mdm.core.storage.sql/src/test/java/com/amalto/core/save/DocumentSaveTest.java
+++ b/org.talend.mdm.core.storage.sql/src/test/java/com/amalto/core/save/DocumentSaveTest.java
@@ -671,6 +671,201 @@ public class DocumentSaveTest extends TestCase {
         assertEquals("1", evaluate(committedElement, "/Person/complex/autoIncrement2"));
     }
 
+    public void testNormalFieldAutoIncrementValueProvided() throws Exception {
+        final MetadataRepository repository = new MetadataRepository();
+        repository.load(DocumentSaveTest.class.getResourceAsStream("metadata24.xsd"));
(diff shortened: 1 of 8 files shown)


org.talend.mdm.core/src/com/amalto/core/save/context/CreateActions.java
```diff
@@ -247,10 +247,15 @@ protected void handleField(FieldMetadata field, boolean doCreate, String current
         // TMDM-3900).
         // Note #2: This code generate values even for non-mandatory fields (but this is expected behavior).
         if (EUUIDCustomType.AUTO_INCREMENT.getName().equalsIgnoreCase(field.getType().getName()) && doCreate) {
-            String conceptName = rootTypeName + "." + field.getPath().replaceAll("/", "."); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
-            String autoIncrementValue = saverSource.nextAutoIncrementId(dataCluster, dataModel, conceptName);
-            actions.add(new FieldUpdateAction(date, source, userName, currentPath, StringUtils.EMPTY, autoIncrementValue,
-                    field, userAction));
+            Accessor accessor = document.createAccessor(currentPath);
+            if (accessor.exist() && !StringUtils.isBlank(accessor.get())) {
```
lpw82gf: For `<fieldA></fieldA>` will be considered as Provided Value or not?

VictorLi-talendbj: In this case, better to consider as not provided value (then it will use system value), otherwise, there are several regression issues from Junit test case.

lpw82gf: Then how about the behavior of bulkload, they must be aligned.

VictorLi-talendbj: For BulkLoad, also changed related code to have same behavior.

org.talend.mdm.core.storage.sql/src/test/java/com/amalto/core/save/DocumentSaveTest.java
```diff
@@ -671,6 +671,201 @@ public void testAutoIncrementForNormalField() throws Exception {
         assertEquals("1", evaluate(committedElement, "/Person/complex/autoIncrement2"));
     }
 
+    public void testNormalFieldAutoIncrementValueProvided() throws Exception {
```
hw-zhu: I suggest to add related jira number on the method comment.

VictorLi-talendbj: Done.

PR #1042, opened 2020-03-16
TMDM-12599 Unable to load the polymorphism record after migration

hw-zhu:
https://jira.talendforge.org/browse/TMDM-12599
Below two REST API.
- GET /data/{containerName}/documents/uniqueIds
-  GET /data/{containerName}/{type}
If there are some inheritance entity in DM, let's say Company and Individual derived from Party, otherwise, the mapped superclass strategy is the simplest approach to mapping an inheritance structure to database tables. It maps each concrete class to its own table. that means that you want to use polymorphic queries, that will get all of record for your use case. 
but now, I only want to get superclass record from the query,  it don't satisfied with our condition for current case.

**What is the new behavior?**
Now, I added an extra parameter **ignoreChild**, that decides whether return the results set which  include the subClass record or not. it has the value **False** by default.

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/src/com/amalto/core/server/api/XmlServer.java
+++ b/org.talend.mdm.core/src/com/amalto/core/server/api/XmlServer.java
@@ -147,6 +147,19 @@ public interface XmlServer {
     String[] getAllDocumentsUniqueID(String clusterName) throws XtentisException;
 
     /**
+     * The list of documents unique ids in a cluster of a particular revision, the second parameter only be used in
+     * Inheratance entity, this is to say if each of entity be mapped into its own table, now we want to get record from
+     * basic table, let's add the parameter ignoreChild with value <b> True </b>.
+     *
+     * @param clusterName The unique ID of the cluster
+     * @param ignoreChild : Determine whether to ignore the sub type record, by default false.
+     * @return the list of document unique IDs
+     */
+    default String[] getAllDocumentsUniqueID(String clusterName, final boolean ignoreChild) throws XtentisException {
+        return new String[0];
+    }
+
+    /**
      * Delete an XML document
      *
      * @param revisionID  The ID of the revision, <code>null</code> for the head
--- a/org.talend.mdm.core/src/com/amalto/core/storage/StorageWrapper.java
+++ b/org.talend.mdm.core/src/com/amalto/core/storage/StorageWrapper.java
@@ -10,6 +10,18 @@
 
 package com.amalto.core.storage;
 
+import static com.amalto.core.query.user.UserQueryBuilder.alias;
+import static com.amalto.core.query.user.UserQueryBuilder.contains;
+import static com.amalto.core.query.user.UserQueryBuilder.eq;
+import static com.amalto.core.query.user.UserQueryBuilder.from;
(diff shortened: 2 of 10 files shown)


org.talend.mdm.query/src/com/amalto/core/query/user/UserQueryBuilder.java
```diff
@@ -227,6 +227,10 @@ public static Condition in(TypedExpression left, TypedExpression right) {
         return new Compare(left, Predicate.IN, right);
     }
 
+    public static Condition in(Expression left, Expression right) {
```
lpw82gf: Can `Condition in(TypedExpression left, TypedExpression right)` meet our requirement?

hw-zhu: class Id directly inherit from interface Expression, so we have to define new method signature.

org.talend.mdm.core/src/com/amalto/core/server/api/XmlServer.java
```diff
@@ -142,9 +142,10 @@ String getDocumentAsString(String clusterName, String uniqueID, String encoding)
      * The list of documents unique ids in a cluster of a particular revision
      *
      * @param clusterName The unique ID of the cluster
+     * @param ignoreChild whether or not to include sub node record
      * @return the list of document unique IDs
      */
-    String[] getAllDocumentsUniqueID(String clusterName) throws XtentisException;
+    String[] getAllDocumentsUniqueID(String clusterName, final boolean ignoreChild) throws XtentisException;
```
VictorLi-talendbj: How can we know when to use `true` or `false` for ignoreChild?

VictorLi-talendbj: Maybe we can add another function, so we can know which function to use by function name, instead of the parameter.

hw-zhu: I suggest using the same method signature, they share the same snippet of code in most cases.

VictorLi-talendbj: Ok, better to to add comment about the new parameter, so we can know when to use `true` or `false`.

org.talend.mdm.core/src/com/amalto/core/storage/StorageWrapper.java
```diff
@@ -291,10 +302,15 @@ public String getDocumentAsString(String clusterName, String uniqueID, String en
 
     @Override
     public String[] getAllDocumentsUniqueID(String clusterName) throws XmlServerException {
-        return getAllDocumentsUniqueID(clusterName, true);
+        return getAllDocumentsUniqueIDDetails(clusterName, true, false);
```
lpw82gf: Advise `findAllDocumentsUniqueID(clusterName, false)`, abut put logic of `getAllDocumentsUniqueIDDetails` in to `findAllDocumentsUniqueID(String clusterName, final boolean ignoreChild)`

PR #1044, opened 2020-03-20
TMDM-14503 Job log file is not produced (upgrade to log4j2)

lpw82gf:
https://jira.talendforge.org/browse/TMDM-14503
- Log not reproduced when run job in MDM through BeforeSaving and Trigger

**What is the new behavior?**
- For Jobs exported from 7.3+ Studio, job runs successfully, log could be produced
- For Jobs exported from 7.2- Studio, job runs successfully, no log produced
- Both Installer and IzPack are OK

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/src/com/amalto/core/jobox/properties/SunOracleStandardPropertiesStrategy.java
+++ b/org.talend.mdm.core/src/com/amalto/core/jobox/properties/SunOracleStandardPropertiesStrategy.java
@@ -80,8 +80,8 @@ class SunOracleStandardPropertiesStrategy implements StandardPropertiesStrategy 
             String mdmRootDir = System.getProperty("mdm.root"); //$NON-NLS-1$
             if (mdmRootDir != null) {
                 properties.put("mdm.root", mdmRootDir); //$NON-NLS-1$
-                File joboxLog4jFile = new File(mdmRootDir + File.separator +  "conf" + File.separator + "log4j-jobox.properties"); //$NON-NLS-1$ //$NON-NLS-2$
-                properties.put("log4j.configuration", joboxLog4jFile.toURI().toURL().toExternalForm()); //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


org.talend.mdm.core/src/com/amalto/core/jobox/properties/SunOracleStandardPropertiesStrategy.java
```diff
@@ -80,8 +80,8 @@ public Properties getStandardProperties() {
             String mdmRootDir = System.getProperty("mdm.root"); //$NON-NLS-1$
             if (mdmRootDir != null) {
                 properties.put("mdm.root", mdmRootDir); //$NON-NLS-1$
-                File joboxLog4jFile = new File(mdmRootDir + File.separator +  "conf" + File.separator + "log4j-jobox.properties"); //$NON-NLS-1$ //$NON-NLS-2$
-                properties.put("log4j.configuration", joboxLog4jFile.toURI().toURL().toExternalForm()); //$NON-NLS-1$
+                File joboxLog4jFile = new File(mdmRootDir + File.separator +  "conf" + File.separator + "log4j-jobox.xml"); //$NON-NLS-1$ //$NON-NLS-2$
+                properties.put("log4j2.configurationFile", joboxLog4jFile.toURI().toURL().toExternalForm()); //$NON-NLS-1$
```
VictorLi-talendbj: What will happen if don't change this file? and only change the content of old file: log4j-jobox.properties, to have format of log4j2.

lpw82gf: Also work as tested, but Pascal think XML would be better.

lpw82gf: MDM itself is using XML format, so align with it should be better.

PR #1057, opened 2020-04-14
TMDM-12599 Unable to load the polymorphism record after migration

hw-zhu:
https://jira.talendforge.org/browse/TMDM-12599
TMDM-14289 don't cover all place which need to add initialized MDMConfiguration object when migration, that caused migrate error.

**What is the new behavior?**
Add new logic code to do the initialize work, that ensure the migration work expected. the piece of code like this:
`    static {
        try {
            MDMConfiguration.createConfiguration("", true);
        } catch (IllegalStateException e) {
            LOGGER.warn("MDMConfiguration had been initialized");
        }
    }`

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/src/com/amalto/core/objects/ObjectPOJOPK.java
+++ b/org.talend.mdm.core/src/com/amalto/core/objects/ObjectPOJOPK.java
@@ -11,47 +11,60 @@ package com.amalto.core.objects;
 
 import java.io.Serializable;
 
+import org.apache.log4j.Logger;
+import org.talend.mdm.commmon.util.core.MDMConfiguration;
+
 import com.amalto.core.util.Util;
 
+public class ObjectPOJOPK implements Serializable {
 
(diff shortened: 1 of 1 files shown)


org.talend.mdm.core/src/com/amalto/core/objects/ObjectPOJOPK.java
```diff
@@ -11,47 +11,60 @@
...
+    String[] ids = null;
 
-	String[] ids = null;
+    static {
+        try {
+            MDMConfiguration.createConfiguration("", true);
+        } catch (IllegalStateException e) {
+            LOGGER.warn("MDMConfiguration had been initialized");
+        }
+    }
```
hw-zhu, line 33: Only add the initialized code, other place changed is causing by format code.

PR #1087, opened 2020-05-21
TMDM-14664 - Update Hazelcast to 3.10.7

coheigea:
Veracode is reporting a high level CVE against Hazelcast 3.7.4. This task is to update to 3.10.7.

### DIFF
--- a/org.talend.mdm.core/pom.xml
+++ b/org.talend.mdm.core/pom.xml
@@ -18,6 +18,7 @@
         <activemq.version>5.15.9</activemq.version>
         <woodstox.core.version>5.3.0</woodstox.core.version>
         <joda.time.version>2.10</joda.time.version>
+        <hazelcast.version>3.10.7</hazelcast.version>
     </properties>
 
     <build>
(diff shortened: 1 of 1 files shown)


org.talend.mdm.core/pom.xml
```diff
@@ -532,12 +533,12 @@
...
             <groupId>com.hazelcast</groupId>
             <artifactId>hazelcast-spring</artifactId>
-            <version>3.7.4</version>
+            <version>${hazelcast.version}</version>
         </dependency>
         <dependency>
             <groupId>com.hazelcast</groupId>
             <artifactId>hazelcast</artifactId>
-            <version>3.7.4</version>
+            <version>${hazelcast.version}</version>
```
hw-zhu, line 541: Hazelcast with version 3.10.7 depends on slf4j library [1.7.25   1.7.30], but the version is slf4j-api(1.6.1) and slf4j-log4j12(1.6.4) in MDM, I think we can upgrade them together. 
@pbabin @lpw82gf

PR #1105, opened 2020-06-15
TMDM-14650 [CVE] - Update Jackson to 2.10.1 (refactor)

lpw82gf:
https://jira.talendforge.org/browse/TMDM-14650
All artificates with groupId start with `com.fasterxml.jackson` have same parent pom as bellow, so define one parameter `jackson.version` would be enough.
```
<parent>
    <groupId>com.fasterxml.jackson</groupId>
    <artifactId>jackson-base</artifactId>
    <version>2.10.1</version>
  </parent>
```

### DIFF
--- a/org.talend.mdm.core/pom.xml
+++ b/org.talend.mdm.core/pom.xml
@@ -15,7 +15,7 @@
     <properties>
         <jackson.version>2.10.1</jackson.version>
         <activemq.version>5.15.9</activemq.version>
-        <woodstox.core.version>5.3.0</woodstox.core.version>
+        <woodstox.core.version>6.0.2</woodstox.core.version>
         <joda.time.version>2.10</joda.time.version>
         <hazelcast.version>3.10.7</hazelcast.version>
     </properties>
@@ -242,6 +242,12 @@
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-joda</artifactId>
             <version>${jackson.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>joda-time</groupId>
+                    <artifactId>joda-time</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.jaxrs</groupId>
--- a/org.talend.mdm.core/pom.xml
+++ b/org.talend.mdm.core/pom.xml
@@ -14,7 +14,6 @@
 
     <properties>
         <jackson.version>2.10.1</jackson.version>
-        <jackson.provider.version>2.10.1</jackson.provider.version>
         <activemq.version>5.15.9</activemq.version>
         <woodstox.core.version>5.3.0</woodstox.core.version>
         <joda.time.version>2.10</joda.time.version>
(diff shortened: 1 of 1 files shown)


org.talend.mdm.core/pom.xml
```diff
@@ -257,32 +271,17 @@
...
-		</dependency>
+            <version>${jackson.version}</version>
+        </dependency>
         <dependency>
-            <groupId>com.fasterxml.jackson.jaxrs</groupId>
-            <artifactId>jackson-jaxrs-json-provider</artifactId>
-            <version>${jackson.provider.version}</version>
+            <groupId>com.fasterxml.woodstox</groupId>
+            <artifactId>woodstox-core</artifactId>
+            <version>${woodstox.core.version}</version>
```
hw-zhu, line 279: woodstox-core depend on stax2-api-4.2.jar, but we have the third dependence in https://github.com/Talend/tmdm-server-se/blob/master/org.talend.mdm.core/pom.xml#L451, the scope is runtime,  actually, the jar exists under  talendmdm/lib.

org.talend.mdm.core/pom.xml
```diff
@@ -232,22 +231,37 @@
...
 	            </exclusion>
             </exclusions>
-            <version>${jackson.provider.version}</version>
+            <version>${jackson.version}</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-xml</artifactId>
-            <version>${jackson.provider.version}</version>
+            <version>${jackson.version}</version>
```
hw-zhu, line 239: As the jar, version 2.10.1 depend on woodstox-core [6.0.2 ~ 6.2.1], but its current version is 5.3.0, if need to upgrade it together?

org.talend.mdm.core/pom.xml
```diff
@@ -232,22 +231,37 @@
...
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-joda</artifactId>
-            <version>${jackson.provider.version}</version>
+            <version>${jackson.version}</version>
```
hw-zhu, line 244: jackson-datatype-jode depend on joda-time, so I suggest exclusion the lower version joda-time.
```
            <exclusions>
                <exclusion>
                    <groupId>joda-time</groupId>
                    <artifactId>joda-time</artifactId>
                </exclusion>
            </exclusions>
```
using directly below 2.10.

org.talend.mdm.core/pom.xml
lpw82gf: Let me upgrade and run the test again.

org.talend.mdm.core/pom.xml
lpw82gf: No need to exclude it explictly, since in the WAR 2.10 will be used already, and 2.9.9 has not CVE issue too.

lpw82gf: But exclude it is also OK.

org.talend.mdm.core/pom.xml
lpw82gf: If the class in a jar not imported directly in mdm's code, we can all add `<scope>runtime</scope>`, stax2-api is just one of them, it is more strict defination. No need to remove.

org.talend.mdm.core/pom.xml
lpw82gf: Upgrade and tested OK.

PR #1149, opened 2020-07-31
TMDM-14573 [6.2.1,7.3.1] random error : Impossible to save itemAGRO_MIR_CLIENT_DonnCom.22531No Support for type 'String3'

hw-zhu:
https://jira.talendforge.org/browse/TMDM-14573
As to REST query `rest/data/DM/query` with condition `"as_of":{"date":"xxxxx"}` in high concurrent request, that will met some weird exception, like `Caused by: java.lang.UnsupportedOperationException: No support for field typed as 'String10'`, and `An exception occurred while closing storage '03891822-3394-4cf7-9bd7-8f86e847fc1c'`, etc.
Tomcat process each request in a separate worker thread, batch processes will cause some concurrent issue, so we have to synchronize individual threads with other parts of your program. and synchronize the access to a shared resource among the multiple threads.

**What is the new behavior?**
Added synchronized block of code to synchronize access to a shared resource in a multi threading environment, plus using some more high level thread-safe collection to avoid above weird issue, so we should replace the existing code with the following code, like `list<String> list = new ArrayList<>(); -> list<String> list = new CopyOnWriteArrayList<>();`.
1. No Shared State
Some state should only be owned by a single thread by using ThreadLocal way.
2. Message Passing
for the threads to communicate. we can implement message passing using a concurrent queue from the package java.util.concurrent. like CopyOnWriteArrayList.
3. Immutable State
Declare its fields as final. this can makes sure that the compiler can check that the fields are in fact immutable.
4. Synchronized Blocks
By putting a lock inside a synchronized block, you make sure that only one thread at a time can execute this section.like
```
                    synchronized (HibernateStorage.class) {
                        SchemaUpdate schemaUpdate = new SchemaUpdate(configuration);
                        schemaUpdate.execute(false, true);
                        exceptions = schemaUpdate.getExceptions();
                    }
```

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/HibernateStorage.java
+++ b/org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/HibernateStorage.java
@@ -1769,7 +1769,11 @@ public class HibernateStorage implements Storage {
             TransactionManager transactionManager = ServerContext.INSTANCE.get().getTransactionManager();
             List<String> transactions = transactionManager.list();
             for (String transaction : transactions) {
-                StorageTransaction storageTransaction = transactionManager.get(transaction).exclude(this);
+                com.amalto.core.storage.transaction.Transaction activeTransaction = transactionManager.get(transaction);
+                if (activeTransaction == null) {
+                    throw new IllegalStateException("Transaction [" + transaction + "] should have been removed from current transactions.\n");
+                }
+                StorageTransaction storageTransaction = activeTransaction.exclude(this);
                 if (storageTransaction != null) {
                     storageTransaction.rollback();
                 }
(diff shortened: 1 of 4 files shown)


org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/hibernate/HibernateStorage.java
```diff
@@ -1769,17 +1769,22 @@ public synchronized void close() {
...
+                    throw new IllegalStateException("Transaction [" + transaction + "] should have been removed from current transactions.\n");
+                }
+                StorageTransaction storageTransaction = activeTransaction.exclude(this);
                 if (storageTransaction != null) {
                     storageTransaction.rollback();
                 }
             }
             // Close Hibernate session
             if (factory != null) {
                 factory.close();
```
hbhong, line 1783: should we close the factory in final block?

hw-zhu: the statement outside has a finally block to clean  all info by destroying object Configuration.

VictorLi-talendbj (approved):

PR #1154, opened 2020-08-07
TMDM-14749 Throw error in server log when modify an item in PROVISIONING

hw-zhu:
https://jira.talendforge.org/browse/TMDM-14749
The behavior is while inserting data into a table without transactions, that caused the insert statements don't be stored to DB.

**What is the new behavior?**
Add begin transaction during business operation, that new data will get inserted into the table as it got rolled back when an error occurred and we have achieved the atomic by using try/catch block.

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/src/com/amalto/core/save/context/DirectWriteContext.java
+++ b/org.talend.mdm.core/src/com/amalto/core/save/context/DirectWriteContext.java
@@ -23,6 +23,7 @@ import com.amalto.core.save.SaverSession;
 import com.amalto.core.save.UserAction;
 import com.amalto.core.server.api.XmlServer;
 import com.amalto.core.util.Util;
+import com.amalto.core.util.XtentisException;
 
 class DirectWriteContext implements DocumentSaverContext {
 
@@ -46,10 +47,17 @@ class DirectWriteContext implements DocumentSaverContext {
 
             @Override
             public void save(SaverSession session, DocumentSaverContext context) {
+                final XmlServer xmlServer = Util.getXmlServerCtrlLocal();
                 try {
-                    final XmlServer xmlServer = Util.getXmlServerCtrlLocal();
+                    xmlServer.start(dataCluster);
                     xmlServer.putDocumentFromString(record, null, dataCluster);
+                    xmlServer.commit(dataCluster);
                 } catch (Exception e) {
+                    try {
+                        xmlServer.rollback(dataCluster);
+                    } catch (XtentisException e1) {
+                        throw new RuntimeException("Error occurred while rolling back a transaction.", e1);
+                    }
                     throw new RuntimeException("Unable to save system record.", e);
                 }
             }


org.talend.mdm.core/src/com/amalto/core/save/context/DirectWriteContext.java
```diff
@@ -46,10 +47,17 @@ public DocumentSaver createSaver() {
...
             @Override
             public void save(SaverSession session, DocumentSaverContext context) {
+                final XmlServer xmlServer = Util.getXmlServerCtrlLocal();
                 try {
-                    final XmlServer xmlServer = Util.getXmlServerCtrlLocal();
+                    xmlServer.start(dataCluster);
                     xmlServer.putDocumentFromString(record, null, dataCluster);
+                    xmlServer.commit(dataCluster);
                 } catch (Exception e) {
+                    try {
```
zbcui, line 56: Should we log the Exception e & XtentisException e1 ?

org.talend.mdm.core/src/com/amalto/core/save/context/DirectWriteContext.java
```diff
@@ -46,10 +47,17 @@ public DocumentSaver createSaver() {
...
+                        xmlServer.rollback(dataCluster);
+                    } catch (XtentisException e1) {
+                        throw new RuntimeException(e1);
```
VictorLi-talendbj, line 59: No log here?

org.talend.mdm.core/src/com/amalto/core/save/context/DirectWriteContext.java
```diff
@@ -46,10 +47,17 @@ public DocumentSaver createSaver() {
...
-                    final XmlServer xmlServer = Util.getXmlServerCtrlLocal();
+                    xmlServer.start(dataCluster);
                     xmlServer.putDocumentFromString(record, null, dataCluster);
+                    xmlServer.commit(dataCluster);
                 } catch (Exception e) {
+                    try {
+                        xmlServer.rollback(dataCluster);
+                    } catch (XtentisException e1) {
+                        throw new RuntimeException("An error occurred when rollback all the statements inside a transaction.", e1);
+                    }
```
htyin, line 60: We always use 'Error occurred...'  without `An`

lpw82gf: `Error occurred while rolling back a transaction.`

org.talend.mdm.core/src/com/amalto/core/save/context/DirectWriteContext.java
lpw82gf: After checking other implementations of DocumentSaverContext, on this level only throw exceptions, and the log will on higher levels. We can keep the same.

org.talend.mdm.core/src/com/amalto/core/save/context/DirectWriteContext.java
hw-zhu: done

PR #1258, opened 2020-11-03
TMDM-14910 Can't save record after adding occurrence using attached customer's datamodel

hw-zhu:
https://jira.talendforge.org/browse/TMDM-14910
If the field is a multiple level polymorphic type in data model, now we don't get the correct metadata type that caused the save error. 

**What is the new behavior?**
I added below snippet of code to fix the case.
```
if (complexTypeMetadata == null) {
     complexTypeMetadata = (ComplexTypeMetadata) field.getType();
}
```

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/src/com/amalto/core/history/accessor/record/DataRecordAccessor.java
+++ b/org.talend.mdm.core/src/com/amalto/core/history/accessor/record/DataRecordAccessor.java
@@ -11,6 +11,7 @@
 package com.amalto.core.history.accessor.record;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.ListIterator;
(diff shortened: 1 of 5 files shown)


org.talend.mdm.core/src/com/amalto/core/history/accessor/record/DataRecordAccessor.java
```diff
@@ -231,18 +232,25 @@ private void create(String specifiedType) {
...
+                                        ComplexTypeMetadata curContainedType = ((ContainedTypeFieldMetadata) field).getContainedType();
+                                        Collection<ComplexTypeMetadata> complexTypeList = curContainedType.getSubTypes();
+                                        for (ComplexTypeMetadata curItem : complexTypeList) {
+                                            if (curItem.getName().equals(specifiedType)) {
+                                                complexTypeMetadata = curItem;
+                                                break;
+                                            }
+                                        }
+                                    }
+                                    if (complexTypeMetadata == null) {
```
VictorLi-talendbj, line 248: How about only adding this part of code to check NULL and without changing above code?

hw-zhu: yes, it also works without any changing, but I think it's convenient to understand it in the present way instead of lambda 
 mode.

VictorLi-talendbj (approved):

PR #1296, opened 2020-12-17
TMDM-14946 High Memory used by MDM Tomcat JVM when calling REST API PUT with "as of "

lpw82gf:
https://jira.talendforge.org/browse/TMDM-14946
High Memory used by MDM Tomcat JVM when calling REST API PUT with "as of "

Under the hood an H2 in-memory database is used to refine the Rest History query prior to serve  the response. A new db is created at each call but is never closed, leading to memory consumption till overwhelming the JVM.

**What is the new behavior?**
The memory is OK when calling same REST API PUT with "as of"

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/inmemory/MemoryStorage.java
+++ b/org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/inmemory/MemoryStorage.java
@@ -57,7 +57,9 @@ public class MemoryStorage extends HibernateStorage {
             Session h2Session = (Session) h2Connection.getSession();
             Database h2Database = h2Session.getDatabase();
             h2Database.shutdownImmediately();
-            LOGGER.info("H2 Database: " + h2Database.getName() + " has been shutted down.");
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("In-memory h2 db '" + h2Database.getName() + "' has been shutted down.");
+            }
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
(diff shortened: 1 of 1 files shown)


org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/inmemory/MemoryStorage.java
```diff
@@ -41,6 +46,23 @@ public void init(DataSourceDefinition dataSourceDefinition) {
...
     }
 
+    @Override
+    public void close() {
+        super.close();
+        try {
+            Class.forName(dataSource.getDriverClassName()).newInstance();
+            Connection connection = DriverManager.getConnection(dataSource.getConnectionURL(), dataSource.getUserName(),
+                    dataSource.getPassword());
+            JdbcConnection h2Connection = (JdbcConnection) connection;
```
hw-zhu, line 56: consider to add checking ```if (connection instanceof JdbcConnection)  {}```

lpw82gf:
> without DB_CLOSE_DELAY=-1, by default, an in-memory database in H2 is discarded after the connection closes
> Then just closing the connection should be enough? (And no H2 specific import here nor a compile dep?)
> 
> `By default, closing the last connection to a database closes the database. For an in-memory database, this means the content is lost. To keep the database open, add ;DB_CLOSE_DELAY=-1 to the database URL. To keep the content of an in-memory database as long as the virtual machine is alive, use jdbc:h2:mem:test;DB_CLOSE_DELAY=-1.`
> 
> But it really depends on the other usage of this class and other changes needed.

Tested, only dropping database files is not enough, `super.close(true)` can also achieve that, but the instances of `org.h2.engine.Database` can't be GC,  because they are strongly referenced in `org.h2.engine.Engine` (see `private static final HashMap<String, Database> DATABASES = New.hashMap();`), we can only shut down the database to totally release the memory.

org.talend.mdm.core.storage.sql/src/main/java/com/amalto/core/storage/inmemory/MemoryStorage.java
lpw82gf: As we explicitly using H2 Driver in `MemoryStorage#init`, so no need to check again the connection type.

pbabin (approved):

hw-zhu (approved):

PR #1356, opened 2021-06-15
TMDM-15085 Improper Neutralization of Directives in Dynamically Eval Code

htyin:
https://jira.talendforge.org/browse/TMDM-15085?

**What is the new behavior?**
Adding more restricted filter for userExpression.

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/org.talend.mdm.core/src/com/amalto/core/util/Util.java
+++ b/org.talend.mdm.core/src/com/amalto/core/util/Util.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -30,6 +31,7 @@ import java.util.Map.Entry;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.MissingResourceException;
 import java.util.Properties;
+import java.util.Set;
 import java.util.UUID;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


org.talend.mdm.core/src/com/amalto/core/util/Util.java
```diff
@@ -1383,4 +1394,19 @@ public static void endTransactionLimit() {
...
     }
+
+	private static boolean isUserContextProperty(String expression) {
+		if (StringUtils.isNotBlank(expression)) {
+			Set<String> userProperties = new HashSet<>();
+			userProperties.add("user_context.id");
+			userProperties.add("user_context.username");
+			userProperties.add("user_context.familyname");
+			userProperties.add("user_context.language");
+			userProperties.add("user_context.properites['']");
```
pbabin, line 1405: Typo here on 'properties'

pbabin (changes requested):

org.talend.mdm.core/src/com/amalto/core/util/Util.java
```diff
@@ -1383,4 +1400,19 @@ public static void endTransactionLimit() {
             TRANSACTION_CURRENT_REQUESTS.decrementAndGet();
         }
     }
+
+	private static boolean isUserContextProperty(String expression) {
```
Gby56, line 1404: Ideal solution would be to return the trusted hardcoded value instead of using the input expression, to break the untrusted data from flowing into the engine... Not sure how complex is the `user_context.properties['....']`

Gby56 (changes requested):
If possible, changing the current sanitizer to return known trusted data only.
Otherwise, the issue will come back in the scan, and after checking Veracode, we don't have annotations to "suppress" the issue in the code directly

org.talend.mdm.core/src/com/amalto/core/util/Util.java
htyin: Properties are defined by user on web UI, can't be hardcoded.

Gby56: Okay, so we'll manually mark the issue as fixed

Gby56 (approved):
Looks good to me !

hw-zhu (approved):

